PR #30, opened 2020-10-09
Display distance and time estimates for routes

### LINKED ISSUE #27: Show approximate route time on search
User testing suggestion: since `pgr_djikstra()` returns the total cost of a given route, we should be able to return an approximate time for a given route by assuming a certain speed ([this blog post suggests 11-18 mph](https://bikecommuterhero.com/whats-the-average-cycling-speed-of-a-bike-commuter/)) and multiplying it by the cost. 

The one snag is that in order to prioritize mellow streets we cheat by reducing the cost in the query:

https://github.com/jeancochrane/mellow-bike-map/blob/6846b7f51272d9aaa09f34d1000b305e37d08876/app/mbm/views.py#L103-L116

So in order to get the true cost we'll need to either figure out how to adjust the `pgr_djisktra()` query to return the unaltered cost, or run an additional query after we get the route to confirm its true cost.

jeancochrane:
## Overview

Use the combined length of the route fragments in meters to display a distance and time estimate when a user searches for a route.

Closes #27.

## Testing instructions

- Ensure tests pass on CI
- If you don't already have a development environment set up, let me know and I'll get you a dump of the database and help you restore it
- Start the development server: `docker-compose up --build`
- Navigate to http://localhost:8000/
- Enter a route and confirm you see a distance and time estimate display in the search box
- Search for another route and confirm the distance and time estimate changes
- Hit "Reset search" and confirm that it clears the distance and time estimate

### DIFF
--- a/app/mbm/views.py
+++ b/app/mbm/views.py
@@ -173,7 +173,7 @@ class Route(APIView):
         distance = f'{formatted_dist} {dist_unit_str}'
 
         # Assume 8mph as a naive guess of speed
-        mi_per_min = 8 / 60
+        mi_per_min = 10 / 60
         time_in_min = dist_in_mi / mi_per_min
         formatted_time = '<1' if time_in_min < 1 else str(round(time_in_min))
         time_unit_str = 'minute' if formatted_time in ['<1', '1'] else 'minutes'
--- a/app/mbm/views.py
+++ b/app/mbm/views.py
@@ -90,7 +90,10 @@ class Route(APIView):
         with connection.cursor() as cursor:
             cursor.execute("""
                 SELECT
-                    way.name, ST_AsGeoJSON(way.the_geom) AS geometry, mellow.type
+                    way.name,
+                    way.length_m,
+                    ST_AsGeoJSON(way.the_geom) AS geometry,
+                    mellow.type
                 FROM pgr_dijkstra(
                     'WITH mellow AS (
                         SELECT DISTINCT(UNNEST(ways)) AS osm_id, type
@@ -131,8 +134,17 @@ class Route(APIView):
             """, [source_vertex_id, target_vertex_id])
             rows = fetchall(cursor)
 
+        # Calculate total distance in miles and time in minutes based on
+        # the total length of the route in meters
+        dist_in_meters = sum(row['length_m'] for row in rows)
+        distance, time = self.format_distance(dist_in_meters)
+
         return {
             'type': 'FeatureCollection',
+            'properties': {
+                'distance': distance,
+                'time': time,
+            },
             'features': [
                 {
                     'type': 'Feature',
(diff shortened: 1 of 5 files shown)


app/mbm/views.py
```diff
@@ -146,6 +158,29 @@ def get_route(self, source_vertex_id, target_vertex_id):
...
+        meters_per_mi = 1609.344
+        dist_in_mi = dist_in_meters / meters_per_mi
+        formatted_dist = round(dist_in_mi, 1)
+        # Don't worry about single-mile case since we always report at least
+        # one decimal (i.e. "1.0 miles")
+        dist_unit_str = 'miles'
+        distance = f'{formatted_dist} {dist_unit_str}'
+
+        # Assume 8mph as a naive guess of speed
+        mi_per_min = 8 / 60
```
kalilsn, line 176: In the issue you mentioned a blog post that estimated 11-18. Anecdotally, that's closer to what I've observed from my own rides. Google seems to be using something around 10mph for a few routes I've looked at around Chicago (although I have no idea if that's constant or what factors they look at).

kalilsn: That said, I think the most important thing is also showing the distance to the user, which you're doing!

kalilsn (commented):
Overall I like this, but I recreated a few routes in google maps for comparison and got pretty different results for distances (e.g. from pilsen to washington park google said 8.4 miles and mbm said 6.4). Any idea what might be causing that? I assume our ways are higher resolution than whatever google is using, but that seems higher than the error I would expect from that.

app/mbm/views.py
jeancochrane: The few blog posts I've read (e.g. [How accurate are Google maps cycling time estimates?](http://www.betterbybicycle.com/2014/09/how-accurate-are-google-maps-cycling.html)) seem to have a consensus that 10mph is the baseline and then they adjust that somewhat mysteriously based on factors like elevation and number of stops. I feel like we have pretty good grounds to bump this up to 10.

jeancochrane:
> Overall I like this, but I recreated a few routes in google maps for comparison and got pretty different results for distances (e.g. from pilsen to washington park google said 8.4 miles and mbm said 6.4). Any idea what might be causing that?

I appreciate this thoroughness! I'm not immediately sure where the error is but the part I'm most uncertain about is how the `length_m` column is being generated. I'm going to do some digging this weekend and see if I can surface the source of the discrepancy.

jeancochrane:
I finally got to the bottom of the incorrect length calculation! The details: it turns out I had forgotten to use `DISTINCT` when joining the route fragments to the `MellowRoute` table in the `get_route()` query, which created duplicate segments in the output in cases where one OSM ID was used for more than one `MellowRoute` (pretty common at the edges of neighborhoods, where the bounding box of one neighborhood often overlaps with another in the route marking interface).

Want to double check to make sure that the routes you tested show the correct distance now @kalilsn?

kalilsn (approved):
Just tested and it looks great!

PR #37, opened 2022-05-15
Lakefront Trail path updates

jackcarter:
### Overview
I noticed that some segments of the lakefront trail were not marked as Paths, especially between Montrose and Fullerton. This commit makes the LFT paths continuous from north trailhead to south trailhead in both directions.

### Demo
**Before:**
image

**After:**
image

Note that this shows non-v2 routing. V2 routing seems to already always choose the LFT. The only benefit for v2 routing is segment color (now shows up as pink instead of blue).


jeancochrane:
Thanks so much for these PRs @jackcarter, I should have some time to review this week but feel free to ping me if you don't hear back!

jeancochrane:
Thanks again for flagging this @jackcarter, and so sorry for the (nearly 2 year 😱 ) delay! I went ahead and fixed the paths in the production app, since unfortunately there's not yet a great system for updating the prod data based on the fixtures stored in the repo. I've got plans to simplify the data refresh system, so hopefully these types of changes will be easier to persist going forward. But in the meantime, I appreciate your contribution!

Issue #35, opened 2021-08-24
Continuous location tracking


jeancochrane:
Poll for location and update it on the map every so often.

PR #40, opened 2022-08-14
Update user location as they move around

kalilsn:
Rough draft for #35

This seems to work pretty well for me testing in Chrome (open devtools, cmd+shift+p -> "Show Sensors", set location to "Other" and change). I tested that it:
	- Updates the map marker as your location changes
	- Doesn't update the map marker if you're not using your current location as the starting point
	- Uses the new location if you click search again

Before merging, I'd like to remove some of the duplication that this creates with the `initAutocomplete` method and make some of these constants more explicit (the zoom level, the "My position" text). This also introduces a bunch of unnecessary `getElementById` calls which I'll fix. I'd also like to refactor so we store the current location in JS rather than writing it to an `<input>` element, although that's a little more involved. Let me know if those sound like okay changes!

I'm also wondering if we need a more explicit way to set the source location to "My position" now that we're actually tracking it. AFAICT the only way to do that right now is by reloading the page (although a quirk of my implementation is that you can now type "My position" into the autocomplete bar and it will then follow you). Perhaps a button next to the input? I'd like to discuss what that interface should look like before I move forward!

### DIFF
--- a/app/mbm/templates/mbm/index.html
+++ b/app/mbm/templates/mbm/index.html
@@ -290,6 +290,10 @@
         ['postal_code', 'short_name', '']
       ]
 
+      const zoomToLocation = (lat, lng) => {
+        map.setView([lat, lng], 14)
+      }
+
       const initAutocomplete = (textElementId, coordElementId, marker) => {
         // Create the autocomplete object
         let autocomplete = new google.maps.places.Autocomplete(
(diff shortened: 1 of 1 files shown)


kalilsn:
Closing this PR and moving this work into a combined PR adding user locations and refactoring the frontend

PR #41, opened 2022-08-21
Let users save locations and factor js into a few modules

### LINKED ISSUE #35: Continuous location tracking

kalilsn:
Opening a draft for some feedback now, but I'll update this again when I finish making the changes that Jean has already asked for:
- [x] Since this disables double click zoom, only enable it if a checkbox is selected (with an explanatory tooltip)
- [x] Change the way we handle geolocation so that:
	- there's always a marker at the user's location
	- if the user uses their position as the source/destination, that marker stays static after a search (while the user location marker continues to move)

I'll also add a new style for the location marker and lightly clean up the styles on the new popup elements I've added.

Aside from UI questions, I'm also interested in feedback on the overall structure here. For example, there's definitely more stuff that could be broken out into its own module to shrink `app.js`, let me know if there's something there I should do. Should the modules be more functional instead of classes? Should the modules be organized differently? Should we implement this like a redux app with state + reducers?

Closes #35


app/mbm/static/js/app.js
```diff
@@ -0,0 +1,345 @@
...
+    // Start the app once the DOM is ready
+    document.addEventListener('DOMContentLoaded', this.start.bind(this))
+  }
+
+  start() {
+    // Create the leaflet map
+    this.map = this.createMap()
+
+    // Store references to DOM elements we'll need
+    const $directionsForm = $('#input-elements')
```
kalilsn, line 24: There was kind of a mixture of jQuery DOM methods and their plain JS equivalents. I didn't make much of an effort to normalize that, but I did prefix jquery vars with `$`

app/mbm/static/js/geolocation.js
```diff
@@ -0,0 +1,36 @@
...
+export default class Geolocation {
+  constructor(app) {
+    this.app = app
+    this.markerName = 'gpslocation'
+    this.startGPSTracking()
+  }
+
+  startGPSTracking() {
+    if (navigator.geolocation) {
+      navigator.geolocation.watchPosition(this.handleGPSPositionUpdate.bind(this))
```
kalilsn, line 14: should we be using `enableHighAccuracy` and/or set a `maximumAge` for some caching? https://developer.mozilla.org/en-US/docs/Web/API/Geolocation/getCurrentPosition#options

app/mbm/static/css/custom.css
```diff
@@ -63,3 +63,61 @@ a {
...
+
+.gps-location-marker-icon:after {
+  content: "";
+  position: absolute;
+  left: 0;
+  top: 0;
+  display: block;
+  width: 100%;
+  height: 100%;
+  background-color: #39d1ff83;
```
jeancochrane, line 98: **[Nitpick, non-blocking]** I wonder if maybe this color is a bit too muted against our background, especially at high zoom levels where it intersects with the route lines? I think a slightly darker shade would provide more contrast, e.g. the color of the marker is very legible and we may even want to copy it for consistency. What do you think?

Screen Shot 2022-08-22 at 1 57 06 PM
Screen Shot 2022-08-22 at 1 59 35 PM

app/mbm/static/js/autocomplete.js
```diff
@@ -0,0 +1,132 @@
...
+
+  const ourOption = $(`
+    <div class="pac-item" tabindex="0">
+      <span class="pac-icon pac-icon-marker"></span>
+      <span class="pac-item-query">${optionText}</span>
+    </div>
+  `).on('mousedown', () => {
+    app.directionsFormElements[markerName].input.value = optionText
+    callback(markerName)
+  })
```
jeancochrane, line 112: **[Nitpick, non-blocking]** As a user I feel like I want the `My position` option to display on focus rather than mousedown 🤔 But I recognize that may be difficult to do with the Google autocomplete library so I'm fine punting it for later, just curious if you agree!

app/mbm/static/js/userlocations.js
```diff
@@ -0,0 +1,134 @@
...
+
+        const popup = L.popup()
+        const sourceButton = $(`<button class="btn btn-primary btn-block">Directions from ${name}</button>`).click(() => {
+            this.app.setSourceLocation(location.lat, location.lng, name)
+            this.map.closePopup(popup)
+        })
+        const targetButton = $(`<button class="btn btn-primary btn-block">Directions to ${name}</button>`).click(() => {
+            this.app.setTargetLocation(location.lat, location.lng, name)
+            this.map.closePopup(popup)
+        })
```
jeancochrane, line 107: **[Nitpick, non-blocking]** From a user perspective, I almost wonder if we want to automatically run the search when the user hits one of these buttons, provided that the other option is already set (e.g. if the user clicks "Directions to home" while the `source` field is populated, we would automatically run the search)? Not a blocking change by any means, but might be a nice UX improvement if you agree!

app/mbm/static/js/userlocations.js
```diff
@@ -0,0 +1,134 @@
...
+
+    // "render" a location to the map by creating a marker at that location,
+    // with a popup containing controls to navigate to/from that marker or to
+    // delete it
+    renderSavedLocation(name, location) {
+        const marker = L.marker([location.lat, location.lng], {
+            title: name,
+            // Make sure these markers sit above the regular source/target markers
+            zIndexOffset: 100,
+        })
```
jeancochrane, line 97: **[Nitpick, non-blocking]** I think it'd be a relatively easy improvement to use a [custom icon](https://leafletjs.com/examples/custom-icons/) like a heart emoji (❤️) to differentiate saved locations from sources/targets; I don't think it needs to block this merge but it'd be nice to do as a fast follow!

jeancochrane (approved):
Really excellent work here! I left a couple suggestions for UX improvements but I don't think any of them needs to block the launch of this feature. Mostly I'm just excited to get it out so I can test the location watching on the ground.

From a high-level code design perspective, I unfortunately don't have much good advice to offer -- I have very little experience with large frontend codebases so I don't really have a good sense of the tradeoffs involved in different design decisions! When I'm working with backend code I typically prefer code that's modularized by feature, and I could see that working here, but above all I'd be down to let you own that decision making if you have opinions about how frontend code should be organized.

app/mbm/static/js/app.js
```diff
@@ -0,0 +1,370 @@
...
+        this.userLocations = null
+      }
+    }
+    this.userLocationsCheckbox.addEventListener('change', this.handleLocationsCheckboxChange)
+
+    this.preferences = getUserPreferences()
+    if (this.preferences.userLocationsEnabled) {
+      this.userLocationsCheckbox.checked = true
+      this.userLocationsCheckbox.dispatchEvent(new Event('change'))
+    }
```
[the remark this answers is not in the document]
kalilsn: >it might be worthwhile for us to eventually factor all the different setup steps into dedicated helper methods to aid in readability

I definitely think we should do this for readability. I started moving towards that but was trying to keep the scope of this from getting too large. Basically I just wanted to do the minimum organization so that these new features don't all have to go in one file, but this App class is way too big and definitely needs to be broken up more.

app/mbm/static/js/autocomplete.js
```diff
@@ -0,0 +1,132 @@
...
+    }
+    app.setMarkerLocation(markerName, lat, lng, addressString)
+  });
+
+  // TODO: This doesn't really need to depend on geolocation. We want to suggest
+  // results all over the city regardless of where the user is located, so we could
+  // instead hardcode a bounds object for chicago city limits. As a bonus, we could
+  // then set the autocomplete to use strictbounds and only return results within
+  // them rather than just biasing towards those results. Alternatively, we could
+  // have the backend generate these bounds based on where it has route definitions.
```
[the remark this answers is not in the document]
kalilsn: Good suggestion, I'll add that to #43

app/mbm/templates/mbm/index.html
```diff
@@ -13,14 +13,21 @@
   <div class="row">
     <div id="controls-container" class="col-12 col-md-3">
       <div class="container-fluid m-auto form-group">
-        <div id="input-elements" class="hideable collapse multi-collapse">
-          <input id="source_text" class="form-control my-2" name="source_text" placeholder="Start address"></input>
```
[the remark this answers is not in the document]
kalilsn: That's correct and it's a little easier to see what I'm doing here in these two commits: https://github.com/jeancochrane/mellow-bike-map/pull/41/commits/5b83114d1bf29b75058482a00dc1f53f8a2e535b and https://github.com/jeancochrane/mellow-bike-map/pull/41/commits/d28991dd61e7ee0e360158e3e40c5a79d9be7586. The first one moves the storage of the actual coordinates that get sent to the backend into our `App` object. The second one removes the `_text` bit from the id because there's no longer a second input to differentiate it from.

app/mbm/static/css/custom.css
```diff
@@ -63,3 +63,61 @@ a {
...
+  0% {
+    transform: scale(0.8);
+  }
+  50% {
+    transform: scale(1);
+  }
+  100% {
+    transform: scale(0.8);
+  }
+}
```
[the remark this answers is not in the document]
kalilsn: This is a good reminder that I need to attribute this because it was 100% stolen from some random tutorial. But it works great!

app/mbm/static/js/autocomplete.js
kalilsn: 100% agree about the behavior, and I actually mention that as a TODO at the top. IDK how to make it happen though :upside_down_face:! I'll make an issue for it as a followup.

This `mousedown` event handler is actually for when you click on the option in the (fake) select menu below the input not for when you focus the input. And bizarrely it doesn't seem to work if I bind to the `click` event.

app/mbm/static/js/userlocations.js
kalilsn: That sounds like a great improvement. I'll open an issue for it!

app/mbm/static/js/userlocations.js
kalilsn: image

This was easy and it does look better, nice call!

kalilsn:
I made two of the changes you asked (darker location marker and hearts for saved locations) and I'll open issues for the rest, then merge this.

I appreciate your thoughts about organization: I also don't know what will work best and I'm just trying things out! Grouping by feature seems reasonable so that's what I'll stick with for now.

app/mbm/static/js/userlocations.js
kalilsn: https://github.com/jeancochrane/mellow-bike-map/issues/44

app/mbm/static/js/autocomplete.js
kalilsn: https://github.com/jeancochrane/mellow-bike-map/issues/45

PR #49, opened 2022-09-10
Add matomo analytics

kalilsn:
Adds matomo analytics

### DIFF
--- a/app/mbm/templates/mbm/base.html
+++ b/app/mbm/templates/mbm/base.html
@@ -41,6 +41,9 @@
 
     <!-- Override this block to add extra stylesheets -->
     {% block extra_css %}{% endblock %}
+
+    <!-- Override this block to add scripts which belong in the <head> -->
+    {% block head_js%}{% endblock %}
   </head>
 
   <body>
--- a/app/mbm/templates/mbm/base.html
+++ b/app/mbm/templates/mbm/base.html
@@ -42,8 +42,21 @@
     <!-- Override this block to add extra stylesheets -->
     {% block extra_css %}{% endblock %}
 
-    <!-- Override this block to add scripts which belong in the <head> -->
-    {% block head_js%}{% endblock %}
+    <!-- Matomo -->
+    <script>
+      var _paq = window._paq = window._paq || [];
+      /* tracker methods like "setCustomDimension" should be called before "trackPageView" */
+      _paq.push(['trackPageView']);
+      _paq.push(['enableLinkTracking']);
+      (function() {
+        var u="//analytics.jeancochrane.com/";
+        _paq.push(['setTrackerUrl', u+'matomo.php']);
+        _paq.push(['setSiteId', '1']);
+        var d=document, g=d.createElement('script'), s=d.getElementsByTagName('script')[0];
+        g.async=true; g.src=u+'matomo.js'; s.parentNode.insertBefore(g,s);
+      })();
+    </script>
+    <!-- End Matomo Code -->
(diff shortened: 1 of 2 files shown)


app/mbm/templates/mbm/index.html
```diff
@@ -9,6 +9,24 @@
   <link rel="stylesheet" href="{% static 'css/spin.css' %}" />
 {% endblock %}
 
+{% block head_js %}
```
jeancochrane, line 12: **[Nitpick, non-blocking]** I wonder if it would make more sense to include this in `base.html` instead of an override in `index.html`, so we can track visits to other pages? Currently the only other page is `About` I think, so it's really not a big deal either way 🤷🏻‍♀️

app/mbm/templates/mbm/base.html
```diff
@@ -41,6 +41,9 @@
 
     <!-- Override this block to add extra stylesheets -->
     {% block extra_css %}{% endblock %}
+
+    <!-- Override this block to add scripts which belong in the <head> -->
+    {% block head_js%}{% endblock %}
```
jeancochrane, line 46: **[Question, non-blocking]** Is there a reason to include this in the `<head>` where it has the potential to block rendering, as opposed to in the `<body>` after the HTML content? It seems like the script is set up to run async so I doubt it matters, but I'm curious!

app/mbm/templates/mbm/index.html
```diff
@@ -9,6 +9,24 @@
...
+  var _paq = window._paq = window._paq || [];
+  /* tracker methods like "setCustomDimension" should be called before "trackPageView" */
+  _paq.push(['trackPageView']);
+  _paq.push(['enableLinkTracking']);
+  (function() {
+    var u="//analytics.jeancochrane.com/";
+    _paq.push(['setTrackerUrl', u+'matomo.php']);
+    _paq.push(['setSiteId', '1']);
+    var d=document, g=d.createElement('script'), s=d.getElementsByTagName('script')[0];
+    g.async=true; g.src=u+'matomo.js'; s.parentNode.insertBefore(g,s);
```
jeancochrane, line 24: **[Question, non-blocking]** Is there a reason that it's necessary for this setup logic to be wrapped in an immediately-executed function expression rather than defined in the body of the enclosing `<script>`? Similarly, I'm curious why this snippet chooses to generate the `<script async>` tag programmatically instead of just defining it in the HTML body?

jeancochrane (approved):
Looks great! I'd love to take a look at the Matomo console to confirm it's tracking things as we expect, but I trust you've tested that out already and I'm happy to take a look at it in production.

app/mbm/templates/mbm/index.html
kalilsn: I forgot about the About page actually. I was meaning to exclude it from the admin pages but there's no real need. I'll put it in the base template.

app/mbm/templates/mbm/base.html
kalilsn: It's what they recommend in matomo, so I can only speculate. My guess is they want it to run before other scripts?

app/mbm/templates/mbm/index.html
kalilsn: The IIFE is to avoid namespace collisions in the global scope with those 1-letter variables, which are pretty common in minified scripts. No clue why they generate the script tag at runtime. I tried googling it and the stackoverflow explanations are very unconvincing!

PR #50, opened 2022-09-13
Disable user locations while we figure out iOS bugs

jeancochrane:
Rather than comment out the user location code, this PR simply removes the checkbox from `index.html` and strips out the section of `App.start()` that previously initialized saved locations and preferences. Once we fix the iOS bugs we can revert the merge commit for this PR to re-enable the feature.


kalilsn (approved):
Sad to see it go but thanks for handling this! I'll come back to it once [kalilsn/vite](https://github.com/jeancochrane/mellow-bike-map/compare/kalilsn/vite) is finished.

PR #56, opened 2024-06-15
Clean up dev environment and update database dependencies

jeancochrane:
This PR consolidates some unnecessary Docker files and updates the database stack to the most recent versions of all the packages. We also update `fixtures/mellowroute.json` to match the most recent dump from prod.

Since the prod app doesn't actually use the development Docker resources or fixtures, these changes shouldn't have any effect on prod.


docker-compose.yml
```diff
@@ -13,13 +13,12 @@ services:
...
-      - ./app:/app
-      - mellow-bike-map-node-modules:/app/node_modules
+      - .:/app
+      - mellow-bike-map-node-modules:/app/app/node_modules
     environment:
       DJANGO_SECRET_KEY: reallysupersecret
       DJANGO_MANAGEPY_MIGRATE: "on"
-    entrypoint: /app/docker-entrypoint.sh
-    command: python manage.py runserver 0.0.0.0:8000
+    command: /app/app/docker-entrypoint.sh python manage.py runserver 0.0.0.0:8000
```
jeancochrane, line 21: Note for posterity: Something weird was happening with `docker-entrypoint.sh` where it couldn't actually read in command overrides as arguments when specifying a new command via the CLI. I never really figured it out, but refactoring things such that `docker-entrypoint.sh` is used in the command rather than the entrypoint fixed the problem, and more closely matches the way I work in development anyway (I don't want to run migrations literally every time I'm running code on the `app` image).

PR #57, opened 2025-02-13
Add Montreal data and page

beamalsky:
As a means of getting reacquainted with this app, I've gone ahead and added a page for Montreal. Wow what a nice project to add to! This worked very smoothly so far.

To test: 

1. Import the new Montreal data. This will take a while: 

`docker compose run --rm -w /app postgres make db/import/montreal.table`

2. Once that's completed, run the app: `docker-compose up`
3. Visit the index page and make sure nothing is broken: http://localhost:8000/
4. Visit the ✨ new Montreal page ✨ and check it out: http://localhost:8000/montreal

I'll leave some more notes below. This is my first time looking at a Django app in like 3 years so definitely open to code organization/style changes.

### DIFF
--- a/app/mbm/static/js/autocomplete.js
+++ b/app/mbm/static/js/autocomplete.js
@@ -11,7 +11,7 @@ const canonicalComponents = [
 const initAutocomplete = (inputElement, markerName, app) => {
   // Create the autocomplete object
   let autocomplete = new google.maps.places.Autocomplete(
-    inputElement, { componentRestrictions: { country: "us" } }
+    inputElement, { componentRestrictions: { country: ["us", "ca"], } }
   )
   // Avoid paying for data that you don't need by restricting the set of
   // place fields that are returned to just the address components.
--- a/app/mbm/templates/mbm/montreal.html
+++ b/app/mbm/templates/mbm/montreal.html
@@ -28,7 +28,7 @@
           </div>
           <div class="form-group">
             <div>
-              <input type="checkbox" id="enable-v2" name="enable-v2" />
+              <input type="checkbox" id="enable-v2" name="enable-v2" checked />
               <label for="enable-v2" style="padding-left: 5px">Use v2 routing</label>
               <i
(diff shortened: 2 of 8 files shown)


app/mbm/static/js/autocomplete.js
```diff
@@ -11,7 +11,7 @@ const canonicalComponents = [
 const initAutocomplete = (inputElement, markerName, app) => {
   // Create the autocomplete object
   let autocomplete = new google.maps.places.Autocomplete(
-    inputElement, { componentRestrictions: { country: "us" } }
+    inputElement, { componentRestrictions: { country: ["us", "ca"], } }
```
beamalsky, line 14: This means the autocomplete options will include both US and Canada results for both the Chicago and Montreal pages. I think this is fine because they're sorted by proximity to the user anyway, but if we wanted to we could only show US results on the Chicago page and Montreal results on the Canada page.

app/mbm/templates/mbm/montreal.html
```diff
@@ -0,0 +1,110 @@
...
+            <label for="target" class="sr-only">End address</label>
+            <input id="target" class="form-control my-2" name="target" placeholder="End address"></input>
+          </div>
+          <div class="form-group">
+            <button id="submit" type="submit" class="btn btn-primary btn-block mt-1">Search</button>
+            <button id="reset-search" type="reset" class="btn btn-secondary btn-block mb-2">Reset search</button>
+          </div>
+          <div class="form-group">
+            <div>
+              <input type="checkbox" id="enable-v2" name="enable-v2" checked />
```
beamalsky, line 31: This new file is mostly copied from `index.html`, other than turning on `enable-v2` by default here and passing the city name to `new App()`.

app/mbm/static/js/autocomplete.js
beamalsky: This thing: 

CleanShot 2025-02-12 at 19 02 55@2x

PR #58, opened 2025-10-15
Update Node from 12.x to 22.x

wrangtangle:
Node 12 is EOLed and didn't install when I ran the Dockerfile. The `npm install` line failed with a "npm not found" error. Updating to active LTS version 22 works.


jeancochrane (approved):
Just rebuilt the app container and confirmed that everything works fine. Thanks!

PR #60, opened 2025-10-15
Bound Google Maps search to Chicagoland

wrangtangle:
Results from Ohio and Texas show up in Google Maps autocomplete. Bound Google Maps search to this polygon (eastern boundary expanded per @jeancochrane's suggestion):

image

For now I've left alone the other code (and relevant TODO comment) related to geolocation and autocomplete, but LMK if you have thoughts on those things.

### DIFF
--- a/app/mbm/static/js/autocomplete.js
+++ b/app/mbm/static/js/autocomplete.js
@@ -12,7 +12,7 @@ const initAutocomplete = (inputElement, markerName, app) => {
   // Define bounds for Chicagoland
   const chicagoBounds = new google.maps.LatLngBounds(
     new google.maps.LatLng(41.45, -88.3),
-    new google.maps.LatLng(42.2, -87.5)
+    new google.maps.LatLng(42.2, -87.3)
   )
 
   // Create the autocomplete object with Chicagoland bounds and strict bounds enforcement


app/mbm/static/js/autocomplete.js
```diff
@@ -9,9 +9,19 @@ const canonicalComponents = [
 
 
 const initAutocomplete = (inputElement, markerName, app) => {
-  // Create the autocomplete object
+  // Define bounds for Chicagoland
+  const chicagoBounds = new google.maps.LatLngBounds(
+    new google.maps.LatLng(41.45, -88.3),
+    new google.maps.LatLng(42.2, -87.5)
```
jeancochrane: **[Nitpick, optional]** I would advocate we expand the easternmost boundary of this box ever so slightly, since it currently runs right through Whiting, which is very close to the eastern limit of streets that are mapped in the application:

Image
Image

Two tenths of a degree should get us out to around Gary, which feels appropriate to me as an eastern boundary, though that's ultimately subjective:

```suggestion
    new google.maps.LatLng(42.2, -87.3)
```

Any concerns with that change?

jeancochrane (approved):
This seems to work well! Let me know what you think about the eastern boundary and I'll merge this in once we resolve the thread.

PR #61, opened 2025-10-15
Store source and target addresses and coordinates in query params

wrangtangle:
Hitting the search button updates the URL to store source and target addresses and coordinates in query params: `sourceAddress`, `sourceCoordinates`, `targetAddress`, `targetCoordinates`.

A URL is valid with any combination of these four query params. Loading a URL with at least one source and one target param runs search. If only one side is present, a marker location is set and search is not run. Coordinates params in the URL eliminate the need to geocode addresses to retrieve their coordinates.

Why do this? There's no way to share directions right now, which would be great from a user perspective. From a development perspective, it'd be great to be able to reload the page with locations already loaded to test changes.

### DIFF
--- a/app/tests/app.start.test.mjs
+++ b/app/tests/app.start.test.mjs
@@ -123,6 +123,21 @@ test('shows coordinate strings when no address is provided', async () => {
   assert.equal(getSubmitCount(), 1)
 })
 
+test('does not geocode when coordinates provided in query params', async () => {
+  const { app } = createApp()
+  let geocodeCount = 0
+  app.geocodeAddress = () => {
+    geocodeCount += 1
+    return Promise.resolve({ lat: 1, lng: 2 })
+  }
+
+  await app.applyInitialQueryParams('?sourceCoordinates=41.9,-87.6&targetCoordinates=42.0,-87.7')
+
+  assert.equal(geocodeCount, 0)
+  assert.equal(app.sourceLocation, '41.9,-87.6')
+  assert.equal(app.targetLocation, '42,-87.7')
+})
+
 test('skips submission when geocoding fails', async () => {
   const { app, getSubmitCount } = createApp({ fromAddress: 'Start', toAddress: 'End' })
   const alerts = []


app/mbm/static/js/app.js
```diff
@@ -152,6 +157,102 @@ export default class App {
...
+      const targetDisplay = targetAddress || targetCoordsParam
+      this.setTargetLocation(targetCoordsFromUrl.lat, targetCoordsFromUrl.lng, targetDisplay)
+    }
+
+    const hasSourceInput = Boolean(sourceAddress) || Boolean(sourceCoordsFromUrl)
+    const hasTargetInput = Boolean(targetAddress) || Boolean(targetCoordsFromUrl)
+    if ((hasSourceInput && !hasTargetInput) || (!hasSourceInput && hasTargetInput)) {
+      alert("Can't load route: start and end locations each require an address or coordinates.")
+      return false
+    }
```
jeancochrane: **[Suggestion, optional]** My instinct is that it's actually fine if one of the source or target is missing -- if that's the case, we can just prepopulate the form and the position for the param that's present, and skip querying the route (which is what the rest of the code already seems to do). From my perspective, that feels like it would be a less obtrusive flow than raising an error via an alert, and the user will still be able to see that one of the source or target is missing and they'll have an opportunity to fix it. That's a weakly-held opinion though, I'm open to hearing a case that a louder error is a better user flow!

app/tests/app.start.test.mjs
```diff
@@ -0,0 +1,206 @@
...
+  const params = new URLSearchParams(lastUrl.split('?')[1] || '')
+  assert.equal(params.get('sourceAddress'), 'Start Address')
+  assert.equal(params.get('targetAddress'), 'End Address')
+  assert.equal(params.get('sourceCoordinates'), '1,2')
+  assert.equal(params.get('targetCoordinates'), '3,4')
+})
+
+test('manual search preserves unrelated query parameters', async () => {
+  const historyCalls = []
+  global.window.location.search = '?utm_source=chatgpt.com'
```
jeancochrane: **[Nitpick, optional]** This is a petty opinion, but I'd prefer not to reference chatgpt:

```suggestion
  global.window.location.search = '?utm_source=duckduckgo.com'
```

README.md
```diff
@@ -37,6 +37,15 @@ docker-compose up
 
 The app will be available on http://localhost:8000.
 
+### Testing
+
+To run frontend tests, from project root:
+
+```
+cd app
+npm test
```
jeancochrane: **[Nitpick, optional]** To match the rest of the instructions in this doc, can we provide Docker commands here too?

jeancochrane (commented):
This is getting close to ready!

> A UX question that this change raises: what should we put in the search inputs when loading a URL in which the source or target has only coords and no address?

I agree with your instinct here (populate the search box with coordinates and use those as the basis for search). Feels like the simplest and most intuitive way to handle this, because the code resolves addresses to coordinates internally anyway.

> Also, thoughts on a share button to copy a link to the route? Could handle in a different PR in you think it's a good idea.

I'm down for this! I don't have a great sense of where it should live on the page, but I'm open to a proposal if you want to draft something. It can be in this PR or in a follow-up, I'm fine either way.

> Also added some tests for the FE logic (the idiomaticity of which tests I unfortunately definitely cannot vouch for).

I am also not totally sure how idiomatic these tests are, but they made sense to me as someone who does very little JavaScript development, so I'm fine bringing them in -- if they're not idiomatic, then at least they're easy enough to read.

app/mbm/static/js/app.js
```diff
@@ -265,6 +430,17 @@ export default class App {
...
+      // Update URL with from/to addresses
+      // Use the stored address strings, or fall back to the input values
+      const fromAddr = this.sourceAddressString
+      const toAddr = this.targetAddressString
+      const sourceCoords = this.sourceLocation
+      const targetCoords = this.targetLocation
+
+      if (fromAddr && toAddr) {
+        this.setRouteQueryParams(fromAddr, toAddr, sourceCoords, targetCoords)
+      }
```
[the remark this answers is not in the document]
wrangtangle: I don't totally understand what you're suggesting here, can you explain more?

wrangtangle:
> I agree with your instinct here (populate the search box with coordinates and use those as the basis for search). Feels like the simplest and most intuitive way to handle this, because the code resolves addresses to coordinates internally anyway.

Implemented this.

> I'm down for this! I don't have a great sense of where it should live on the page, but I'm open to a proposal if you want to draft something. It can be in this PR or in a follow-up, I'm fine either way.

I think follow-up PR is probably easiest since there'll probably be some iterations around UI.

I realized that there's a missing piece here: the <title> is always `Home | Mellow Bike Map`, but now that we are changing the URL, we ought to set the URL on page load and when a new search is done, so stuff like `49 West Chicago Avenue, Chicago, IL 60654 to 59 West Chicago Avenue, Chicago, IL 60654 | Mellow Bike Map` and `41.89622560000001,-87.62874169999999 | Mellow Bike Map`. I'd suggest I do this in a follow-up since it's an improvement and this PR keeps the current behavior, and this PR is already longish, but LMK what you think.

app/mbm/static/js/app.js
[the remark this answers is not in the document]
jeancochrane: I was just noticing that we already assign `this.sourceLocation` and `this.targetLocation` to variables shortly before this block, so there's not necessarily a reason to reassign them to `sourceCoords` and `targetCoords`:

https://github.com/jeancochrane/mellow-bike-map/blob/1cd8557a4f124a8819c74cb590b7a17ab5d7c77f/app/mbm/static/js/app.js#L260-L261

But now that I'm rereading my comment a few days later, I'm realizing that this was a pointless nitpick, so I don't mind leaving it as-is!

PR #67, opened 2025-10-24
Toggle visibility of street calmness levels

wrangtangle:
Click street calmness levels in the map legend to toggle their visibility. All calmness levels on a plotted route remain visible even when they have been toggled off.

image

Why do this? Really because it makes it easier to look at just the mellow streets, which at least I as a user care about seeing the most on here.

QA steps I tried:
- Toggle all the calmness levels on/off
- Same after searching for a route
- Search for a new route and toggle

### DIFF
--- a/app/mbm/static/js/app.js
+++ b/app/mbm/static/js/app.js
@@ -9,9 +9,13 @@ export default class App {
     this.routeListUrl = routeListUrl
     this.routeUrl = routeUrl
 
+    // The layer that displays the route between the source and target locations
     this.directionsRouteLayer = null
+
+    // The layer that displays the calm routes on the map: "off-street bike paths", "mellow streets", and "main streets, often with bike lanes"
     this.calmRoutesLayer = null
     this.calmRoutesData = null
+
     this.markers = { 'source': null, 'target': null }
 
     this.routeTypes = {


app/mbm/static/js/app.js
```diff
@@ -182,13 +191,33 @@ export default class App {
...
+        const color = this.getLineColor(type)
         const description = routeType[1]
-        div.innerHTML += `<i style="background:${color}"></i>${description}`
-        if (routeType !== routeTypes[routeTypes.length - 1]) {
-          div.innerHTML += '<br>'
-        }
+        
+        // Create a container for each legend item
+        const item = L.DomUtil.create('div', 'legend-item', div)
+        item.style.cursor = 'pointer'
```
jeancochrane: **[Nitpick, optional]** Since this attribute is constant, couldn't we just define it on the `legend-item` styles in `custom.css` instead?

app/mbm/static/js/app.js
```diff
@@ -203,13 +232,40 @@ export default class App {
...
+  }
+
+  // Reload all routes with current filters
+  reloadCalmRoutes() {
+    if (this.calmRoutesLayer) {
+      this.map.removeLayer(this.calmRoutesLayer)
+      this.calmRoutesLayer = null
+    }
+    this.loadCalmRoutes()
+  }
```
jeancochrane: **[Suggestion, blocking]** This works, and it's nice that it's a relatively simple change, but I don't like that it requires an additional network request to filter data that we already have. Unless it's prohibitively complex to filter the existing routes by type, I think we should do that rather than re-request them every time we want to filter them.

jeancochrane (changes requested):
I really like this feature! A couple thoughts below on cleaning up the approach.

app/mbm/static/js/app.js
wrangtangle: Good point, I've updated this to prevent additional network requests.

app/mbm/static/js/app.js
wrangtangle: Definitely, have moved this.

app/mbm/static/js/app.js
```diff
@@ -9,10 +9,16 @@ export default class App {
     this.routeListUrl = routeListUrl
     this.routeUrl = routeUrl
 
-    this.routeLayer = null
-    this.allRoutesLayer = null
+    this.directionsRouteLayer = null
+    this.calmRoutesLayer = null
```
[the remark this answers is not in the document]
jeancochrane: Ah, got it! Yeah, I agree that the term "route" is currently overloaded. It's confusing and we should find better terminology.
I'm going to stew on it a bit more to try to figure out a categorization scheme that I like better, but in the meantime, I'm happy to switch to `calmRoutesLayer`. Let's just drop a quick comment above these two variable assignments so that my future self can get a quick reminder about what each one means.

jeancochrane (commented):
I think this is good to go! Any last changes you were planning, or am I good to merge?

PR #68, opened 2025-10-25
Move settings from sidebar to settings menu

wrangtangle:
Move settings from sidebar into a dropdown menu in navbar.

Why? This frees up real estate to put turn-by-turn directions in sidebar on desktop. (I was thinking that on desktop, turn-by-turn directions could start out in the sidebar, and maybe there'd be a button to click to expand them. That would let us show them without altering the rest of the current UI. On mobile, turn-by-turn directions could just be below map.) However, if we don't want to go that direction for the UI of turn-by-turn directions, then this is not needed.

I've removed the search and legend toggle buttons from the desktop view, since they weren't needed there, but left them in the mobile view.

## Before

image

## After

image

image

### DIFF
--- a/app/mbm/static/js/app.js
+++ b/app/mbm/static/js/app.js
@@ -54,18 +54,87 @@ export default class App {
 
     this.gpsLocationString = 'My position'
 
-    const isMobileScreen = $(window).outerWidth() <= 768
+    const getIsMobileScreen = () => $(window).outerWidth() <= 768
+    let previousIsMobileScreen = null
 
-    // Make sure the map always fits the full height of the screen
-    $(window).resize(() => {
+    this.$hideSearch = $('#hide')
+    this.$hideLegend = $('#hide-legend')
+
+    const setSearchToggleState = (state) => {
+      if (!this.$hideSearch.length) { return }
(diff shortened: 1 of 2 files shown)


jeancochrane (commented):
I did a quick test drive of this change and I like it! I don't actually know if we want to keep the v2 routing and saved location options -- both were experiments that didn't really pan out, and it might make our lives easier to reduce maintenance burden by stripping them out -- but that doesn't need to happen in this PR, particularly since there's not yet any other features to replace them with.

I'd like to read through the code in more detail before I approve, and I should have time for that on Thursday or Friday.

wrangtangle:
> I don't actually know if we want to keep the v2 routing and saved location options -- both were experiments that didn't really pan out, and it might make our lives easier to reduce maintenance burden by stripping them out -- but that doesn't need to happen in this PR, particularly since there's not yet any other features to replace them with.

Makes sense! 

Saved locations adds an important bit of functionality right now IMO - it lets you generate a route to/from points without an address (e.g. on park paths). This was very useful for me in testing #73. There are other ways to get to the same place, though (and now we support routing to/from coords).

Issue #64, opened 2025-10-17
Routing onto sidewalks


wrangtangle:
From `45 West Chicago Avenue, Chicago, IL 60654` to `East 56th Place, Chicago, IL 60637` sends you onto an osm_way tagged [`footway=sidewalk`](https://wiki.openstreetmap.org/wiki/Tag:footway%3Dsidewalk), and from what I can tell from Google Maps imagery, it does seem to be a normal sidewalk.

<img width="314" height="335" alt="Image" src="https://github.com/user-attachments/assets/e31313af-2da0-442f-9e54-46700180ea38" />

<img width="533" height="594" alt="Image" src="https://github.com/user-attachments/assets/e2eafe08-75a7-4838-b15e-35e8ca19dfa4" />

Not sure if everything tagged `footway=sidewalk` is undesirable to route onto in Chicago, haven't looked into it yet, but this does not seem to be ideal routing here.

PR #70, opened 2025-11-07
Penalize routing onto sidewalks

wrangtangle:
~~Prevent~~ penalize routing onto sidewalks when route cost is calculated. 

Initially assumed that we could completely prevent routing onto sidewalks (example of problem: https://github.com/jeancochrane/mellow-bike-map/issues/64), but this breaks too many routes.

### DIFF
--- a/app/mbm/static/css/custom.css
+++ b/app/mbm/static/css/custom.css
@@ -135,20 +135,3 @@ a {
     transform: scale(0.8);
   }
 }
-
-.route-debug {
-  background-color: #fff9e6;
-  border: 1px solid #f0d860;
-  border-radius: 4px;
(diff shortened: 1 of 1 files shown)


wrangtangle:
Testing with the script at https://github.com/jeancochrane/mellow-bike-map/pull/70/commits/a6c6770cf1101ede1b642dae2aa7710852828c46, which picks random vertex pairs and tries to route between them with/without routing onto sidewalks allowed, suggests that sidewalks may be needed to route between a large proportion of vertex pairs, potentially >50%.

Penalizing routing onto sidewalks rather than preventing it seems maybe more viable - I'll investigate that.

Issue #69, opened 2025-10-27
Turn-by-turn directions


wrangtangle:
(This is technically a dupe of https://github.com/jeancochrane/mellow-bike-map/issues/7, but wanted to make an issue to map out how I'm thinking about the overall approach and sequence of tasks involved here and get feedback. Will edit this more as understanding of problem improves.)

1. Audit types of ways in use in directions, and see if based on tags or other heuristics we should prevent routing onto some of them. Seems like sometimes routes may go onto undesirable types of ways (e.g., sidewalk: https://github.com/jeancochrane/mellow-bike-map/issues/64). This is less noticeable in the current UI where directions are given entirely as a plotted line on the map, but it could be more problematic in listed turn by turn directions. To the extent that there are problems here, fixing will also improve the quality of directions overall.
1. Classify unnamed ways (https://github.com/jeancochrane/mellow-bike-map/issues/62). Right now all we have the ability to say when a way is unnamed is ~"turn right onto the unnamed street". To the extent possible, we should try to characterize it in some way based on what we can figure out about it - an alley, a park path, etc.
    - Potentially also use landmarks or relative turns ("second right") when there are sequences of unnamed ways: https://github.com/jeancochrane/mellow-bike-map/issues/63
    - Adding smaller maps/images of the turns along with individual turn instructions might help.
1. Create the UI for turn-by-turn directions, building off of https://github.com/jeancochrane/mellow-bike-map/pull/48. My thinking is that on mobile, the directions just show up in a list below the map, and on desktop, the directions first show up in the sidebar, then if you click a button, the size of the map is reduced a bit from 100% height and the list of directions pops over bigger below there. But obviously not married to that approach.
    - If we go with this approach, then we should probably clear space on the sidebar: https://github.com/jeancochrane/mellow-bike-map/pull/68
    - Could also add a print view.

Issue #62, opened 2025-10-16
Classify unnamed streets


wrangtangle:
There are more unnamed `chicago_ways` (710,730, querying the dataset I have) than named (247,216), and the draft of turn by turn directions (#48) gives frequent instructions to turn onto unnamed ways. We should classify as many as possible these in order to be able to say stuff like "take a right turn into the alley" or "go across the crosswalk" (or figure out which of these to skip over) in the turn by turn directions.
 
1. Where possible, give names to long unnamed ways (or rather, long sequences of unnamed ways). They may not be officially named, but we may be able to come up with [definite descriptions](https://en.wikipedia.org/wiki/Definite_description) for them.
1. Make a list of what all the types of unnamed ways are, just by inputting different from/to addresses, looking at the directions, and seeing what unnamed ways represent: alley, crosswalk, bike path, etc.
1. Investigate whether some types of unnamed ways should not be routed onto (see https://github.com/jeancochrane/mellow-bike-map/issues/64).
1. Come up with heuristics for classifying each type of unnamed way. For example, if an unnamed way is parallel to a named way, less than x meters from it, at least y meters long, has/doesn't have particular tags, and meets/doesn't meet various other criteria, potentially could classify it as an alley.     - Heuristics could potentially produce boolean or probabilistic results. Ideally the heuristics are boolean, but if probabilistic, in uncertain cases we could potentially say stuff like "turn right into the alley or pedestrian path to head south" (or phrase it in a way that conveys that we think it's one but there's a chance it might be another). In cases like this, we could potentially try to query the Google Maps API for a street address or business name to use as a landmark in making the turn. Everything with probabilistic heuristics might be a janky user experience, not saying we _should_ do this, just brainstorming.
    - Not in the MVP, but we could potentially expose an interface for users to correct mistaken classifications.
1. Create SQL queries based on the heuristics to run on build and populate an ~`unnamed_way_classification` column. Then we can read this column when giving turn by turn directions. Maybe we also have an ~`unnamed_way_info` column if we want to add bits like "in Humboldt Park" to add useful additional info to the classification.

PR #73, opened 2025-11-14
Turn by turn directions

wrangtangle:
Expands on #48 and https://github.com/jeancochrane/mellow-bike-map/pull/68, and optionally incorporates https://github.com/jeancochrane/mellow-bike-map/pull/61, to implement https://github.com/jeancochrane/mellow-bike-map/issues/69.

## Changes

- Adds turn by turn directions to the UI in the left column on desktop, or below the map on mobile. Individual directions can merge multiple chicago_ways; adding `?debug=true` to the URL adds a panel below each direction that shows the ways it's composed of.

image

_(As you can see from this screenshot, there are still issues with routing onto [sidewalks](https://github.com/jeancochrane/mellow-bike-map/issues/64), which I've opened a separate draft PR for, and with turns onto unnamed streets that could be more specifically identified, though many of the unnamed streets now at least have descriptions.)_

- Describes unnamed ways in directions based on tags:
image

- Stores Chicago park geometries in DB, uses them to label chicago_ways that are in parks on build, and uses this information to describe unnamed ways in directions as being in parks:
image

- Add a script to fuzz turn by turn directions between random vertices to find issues. Instructions on running in readme.

## To do 

- Still doing some refactoring.
- There are still many "unnamed streets" that show up in directions. Fuzz directions more to find more combinations of tags that straightforwardly can resolve into descriptions for unnamed streets.
- I think we could improve the characterization of alleys, e.g. "the alley parallel to x street".
- As discussed in https://github.com/jeancochrane/mellow-bike-map/issues/62, there's more we could do to disambiguate turns on unnamed streets.
- Potentially hide this feature behind a setting: checkbox for "Turn by turn directions (beta)". Or, I'm not sure if there's a staging site? We'd want only a smaller group of users to be seeing this until the kinks are worked out.
- Potentially add little maps to each direction. I was experimenting a bit with doing this with SVG, and it seems potentially doable. I think the minimal version of this that would be useful would represent the previous segment (turning off of), the new segment (turning onto), the roads/paths connected to that vertex, and have labels for anything with a name. I'd expect little maps would be helpful disambiguating turns where several unnamed paths come together.


jeancochrane (commented):
This is really cool! From some preliminary testing, the directions algorithm seems pretty solid, and I think the desktop UX is smooth. The unnamed streets phenomenon doesn't bother me so much, though I'm open to more work on that front if you feel passionate about it.

The mobile experience seems like the hardest design question to me, so I'm curious to hear more about where you imagine that going. I don't think putting content below the map and requiring users to scroll down to it will work very well, because the map is intentionally designed to be full-width with scroll interactions intercepted so that scrolling on the map scrolls the map rather than the page. There are a few ways I can think of getting around this:

1. Do what some web slippy maps do, and make it so that scrolling the map with one finger scrolls the page and scrolling with two fingers scrolls the map
    - Pro: Would require the fewest code changes
    - Con: Confusing UX that makes it harder to use the map, would be particularly annoying for folks who are not interested in turn-by-turn directions
2. Copy mapping apps like Google Maps and introduce a widget that lets you pull up directions by swiping up, then close them by swiping down
    - Pro: Tried-and-true UX that would feel comfortable to users while minimizing use of screen real estate
    - Con: Might be hard to code in a maintainable way given our lack of a robust UI framework 
3. Same as option 2 but slightly simpler: Use a simple button + closable modal for turn-by-turn directions 
    - Pro: UX benefits of 2 with less code
    - Con: Buttons take up a lot of screen real estate on mobile

What do you think? I'd be happy to do a synchronous design session later this week if it would make brainstorming easier, too!

jeancochrane (commented):
Capturing some of our feedback from our in-person UX testing:

  * Remove buffer around nodes when highlighting a route segment (keep the buffer around the ways)
  * Handle navbar sizing on small screens (reduce logo text size, then ultimately remove logo text, leaving just the bike emoji)
  * When clicking on a turn-by-turn route segment, if the directions modal is expanded to full screen, un-expand the modal so that the user can see the segment they've selected
  * When clicking on a turn-by-turn route segment, if the directions modal is _not_ expanded, then the operation to center the map viewport should exclude the portion of the map that is obscured by the directions modal

db/raw/chicago_parks.geojson
jeancochrane, line 1: **[Nitpick, optional]** Can we move this file to `db/raw/` for consistency with the rest of the raw data files?

jeancochrane, line 1: **[Question, blocking]** What's the provenance of this data file? Was it retrieved from the public internet? If so, I think we should either script its download in the Makefile, or include some docs in the Makefile explaining where we got it, in case we ever need to update it.

db/label-parks.sql
```diff
@@ -0,0 +1,70 @@
...
+-- Humanize park names: remove text in parens, title case, add " Park" to the end if not already present
+-- Will be present for park names like "PARK NO. 399", will not be for the majority of parks, like "MAPLEWOOD"
+UPDATE chicago_parks
+SET humanized_name = CASE 
+    WHEN UPPER(TRIM(REGEXP_REPLACE(park, '\s*\([^)]*\)', '', 'g'))) LIKE '%PARK%' 
+    THEN INITCAP(TRIM(REGEXP_REPLACE(park, '\s*\([^)]*\)', '', 'g')))
+    ELSE INITCAP(TRIM(REGEXP_REPLACE(park, '\s*\([^)]*\)', '', 'g'))) || ' Park'
+END
+WHERE park IS NOT NULL AND park != '';
+
```
jeancochrane: **[Suggestion, optional]** I think this logic is a good start! However, based on a quick scan of the inputs and outputs, there are a few edge cases that jump out to me:

1. It's an enormous nitpick, but it looks kind of awkward when this logic capitalizes connector words like "and", "the", and "or", which are usually not capitalized.
2. In most cases, the parenthetical text does appear to be superfluous, and I think it's a good idea to remove it. But I wonder about cases where the parenthetical helps disambiguate multiple parks with identical names? For example, there are three parks named "Jackson Park" in the city, each with a different parenthetical indicating the namesake of that park. For routing purposes, it's may not be a big deal to call all three of these "Jackson Park", though it may be confusing in specific cases where most people incorporate the parenthetical into the park name in casual conversation (specific examples below).
    - A special case of this problem is Ping Tom Park, which is recorded as "TOM (PING) MEML." in the input data, and so gets transformed to "Tom Meml. Park" in the output table -- but I've never heard anyone call it that, and there are no other parks named "Tom Park" that we would need to disambiguate. "Ping Tom Meml. Park" would be much clearer, I think.
    - Thinking about the "Jackson Park" example in more detail: If I were tasked with writing directions, I would probably refer to "Jackson (Andrew) Park" as "Jackson Park", since it's the most significant of the three parks, and I would call the other two parks "Mahalia Jackson Park" and "Robert Jackson Park" . Same thing for "Washington Park", which I associate most strongly with the largest of the Washington Parks (that would be "Washington (George) Park"), while in contrast I would notate "Washington (Harold) Park" as "Harold Washington Park". So maybe the rule here should be that, when multiple parks share the same surname, the most significant of those parks should take the form "<Surname> Park", while the smaller parks should take the form "<Parenthetical> <Surname> Park".
3. Some of the abbreviations in here might be confusing, like "Cultl. Ctr." for "Cultural Center".

Point 1 is probably achievable in code, but I wonder if points 2 and 3 call for a crosswalk for park names with specific, non-generalizable name requirements? I would be happy to draft that crosswalk if you agree that it would be useful. But I'm also curious to hear if you think this whole thing is worth spending more time on at all; perhaps it only matters for parks that we use for routes, which is nowhere near the majority of parks.

app/mbm/management/commands/generate_directions.py
```diff
@@ -0,0 +1,225 @@
...
+        )
+        rows = fetchall(cursor)
+    if not rows:
+        return None, None, None, None
+    return (
+        rows[0]["component"],
+        rows[0]["component_node_count"],
+        rows[0]["total_components"],
+        rows[0]["total_nodes"],
+    )
```
jeancochrane, line 69: **[Suggestion, optional]** It's probably fine since this is a management command and not module code, but situations like these where we're returning a bunch of values in a tuple often call for a [dataclass](https://docs.python.org/3/library/dataclasses.html).

app/mbm/static/js/turnbyturn/displayDirections.js
```diff
@@ -0,0 +1,222 @@
...
+  positionDirectionsContainer(app)
+
+  // On mobile, reset the explicit show state when new directions are displayed
+  const isMobileScreen = $(window).outerWidth() <= 768
+  if (isMobileScreen) {
+    app.mobileDirectionsShown = false
+  }
+}
+
+export function positionDirectionsContainer(app) {
```
jeancochrane, line 100: **[Suggestion, optional]** This is a lot of logic to execute on every window resize. Plus, it seems to be fundamentally display logic. I wonder if there's a way we could accomplish it with CCS instead of JS?

README.md
jeancochrane, line 1: **[Nitpick, optional]** I think we may need to add import instructions for loading parks, right?

app/mbm/urls.py
```diff
@@ -24,6 +24,7 @@
     path('about/', views.About.as_view(), name='about'),
     path('api/route/', views.Route.as_view(), name='route'),
     path('api/routes/', cache_page(60 * 60 * 24)(views.RouteList.as_view()), name='route-list'),
+    path('api/parks/', cache_page(60 * 60 * 24)(views.ParkBoundaries.as_view()), name='parks'),
```
jeancochrane: **[Question, non-blocking]** I think we discussed this during our last session, but remind me: Is there a reason to keep this API endpoint in?

app/mbm/routing.py
```diff
@@ -0,0 +1,167 @@
...
+
+CYCLEWAY_TAG_IDS = (
+    101,  # cycleway:track
+    201,  # cycleway:right:track
+    301,  # cycleway:left:track
+    501,  # highway:cycleway
+)
+
+
+def get_nearest_vertex_id(coord: List[float]) -> int:
```
jeancochrane: TK: Don't move this to `routing.py` just yet, so that we can see the diff easier

app/mbm/directions.py
```diff
@@ -0,0 +1,245 @@
...
+    elif osm_tags.get('highway') == 'cycleway':
+        description = 'a bike path'
+    elif osm_tags.get('bicycle') == 'designated':
+        description = 'a bike path'
+    elif osm_tags.get('highway') == 'pedestrian':
+        description = 'a pedestrian path'
+    elif osm_tags.get('highway') == 'footway' or osm_tags.get('footway') == 'sidewalk':
+        description = 'a sidewalk'
+    elif osm_tags.get('highway') == 'pedestrian' and osm_tags.get('bicycle') == 'yes':
+        description = 'a mixed-use path'
```
jeancochrane, line 158: **[Nitpick, non-blocking]** This conditional is a no-op since we already check `highway == 'pedestrian'` on line 154 above. Not a huge deal, but if we want to preserve this condition, we should move it above the check on line 154 so it takes priority.

app/mbm/directions.py
```diff
@@ -0,0 +1,245 @@
...
+    is_slight_turn = (maneuver == 'Turn slightly to the left' or 
+                     maneuver == 'Turn slightly to the right' or maneuver == 'Continue')
+    is_named_street = bool(name)
+    same_named_street = new_segment_effective_name == previous_segment_effective_name
+    if is_slight_turn and is_named_street and same_named_street:
+        return True
+    
+    # Merge continues on unnamed streets
+    if maneuver == 'Continue' and not is_named_street:
+        return True
```
jeancochrane, line 214: **[Question, non-blocking]** This is always going to be true for the first segment whenever it is not a named street. Does that pose a problem for us? I notice the only similar test case has a named street as well (`test_first_segment_no_previous_heading_should_not_merge`). Might be worth testing the unnamed case to make sure the output is what we expect.

app/mbm/views.py
```diff
@@ -87,80 +84,13 @@ def get_coord_from_request(self, request, key):
...
+        directions = directions_list(features)
 
-        # Calculate total distance in miles and time in minutes based on
-        # the total length of the route in meters
-        dist_in_meters = sum(row['length_m'] for row in rows)
-        distance, time = self.format_distance(dist_in_meters)
-        major_streets = self.get_major_streets(rows, dist_in_meters)
+        # need to adapt to not take rows
+        # major_streets = self.get_major_streets(rows, dist_in_meters)
+        major_streets = []
```
jeancochrane, line 93: **[Question, blocking]** What's going on here? I'm not sure why this code is commented out.

app/mbm/directions.py
```diff
@@ -0,0 +1,245 @@
...
+    directions: List[Direction] = []
+    previous_heading = None
+    previous_effective_name = None
+    
+    for i, feature in enumerate(features):
+        props: RouteProperties = feature['properties']
+        name = props.get('name')
+        heading: float = props.get('heading', 0.0) or 0.0
+        distance: float = props.get('distance', 0)
+        route_type: Optional[str] = props.get('type')
```
jeancochrane: **[Question, non-blocking]** Are we using this variable anywhere?

app/mbm/management/commands/generate_directions.py
```diff
@@ -0,0 +1,225 @@
...
+                self.stdout.write(
+                    "Vertices are in different connected components."
+                )
+            raise CommandError(
+                "No route found between source and target coordinates."
+            )
+        # Print basic route information
+        total_distance = sum(
+            feature.get("properties", {}).get("length_m", 0) for feature in features
+        )
```
jeancochrane, line 194: **[Question, non-blocking]** Isn't `calculate_route()` using `distance` as this property name instead of `length_m`? ([source](https://github.com/wrangtangle/mellow-bike-map/blob/98f9c7aa7e97db10d4cbf7a06be8b5bf25031957/app/mbm/routing.py#L124))

```python
                'distance': row['length_m'],
```

app/mbm/directions.py
```diff
@@ -0,0 +1,245 @@
...
+        return int(value + 0.5)
+
+    if miles < 0.09:
+        feet = round_half_up(meters / meters_per_foot)
+        unit = 'foot' if feet == 1 else 'feet'
+        return f"{feet} {unit}"
+
+    rounded_miles = round_half_up(miles * 10) / 10
+    unit = 'mile' if rounded_miles == 1 else 'miles'
+    return f"{rounded_miles} {unit}"
```
jeancochrane, line 129: **[Question, non-blocking]** This seems like a dupe of `mbm.routing._format_distance()`, albeit with slightly different logic. Do they really need to be two different functions? If so, is there a way we can distinguish their function names to make the difference clearer?

app/mbm/static/js/turnbyturn/displayDirections.js
```diff
@@ -0,0 +1,222 @@
...
+
+    const icon = getDirectionIcon(maneuver, color)
+
+    const directionText = direction.directionText || ''
+
+    let listItemHtml = `<li class="direction-item" data-direction-index="${index}" data-color="${color}"><span class="direction-icon-wrapper">${icon}</span><span class="direction-text">${directionText}`
+
+    listItemHtml += `</span></li>`
+
+    $directionsList.append(listItemHtml)
```
jeancochrane, line 22: **[Nitpick, optional]** I'm not terribly concerned about it, but I think it's technically possible to insert malicious code into an OSM street name in order to get it templated in here via `directionText`. Whether or not that's a realistic attack vector is a separate question, but it's easy enough to use the JQuery `text()` API to sanitize the input (code not tested):

```suggestion
const $li = $('<li>', {
    class: 'direction-item',
    'data-direction-index': index,
    'data-color': color,
})

// icon is a static SVG string we control, so innerHTML is fine here
const $iconWrapper = $('<span>', { class: 'direction-icon-wrapper' }).html(icon)

const $textWrapper = $('<span>', { class: 'direction-text' })
$textWrapper.text(directionText)

$li.append($iconWrapper).append($textWrapper)
$directionsList.append($li)
```

app/mbm/static/js/app.js
```diff
@@ -688,7 +872,120 @@ export default class App {
...
+        return {
+          weight: 8,
+          color: color,
+          opacity: 1
+        }
+      },
+      onEachFeature: function (feature, layer) {
+        // Bind popup to allow tooltips on highlighted segments
+        layer.bindPopup(
+          `<strong>Name:</strong> ${feature.properties.name}<br>` +
```
jeancochrane, line 989: **[Nitpick, optional]** Here we are also potentially directly rendering untrusted input from OSM, I think. Not totally sure the best way to do so but we should probably escape it if we want to be maximally defensive.

jeancochrane (commented):
Looking great! Lots of small thoughts, but nothing disqualifying just yet.

jeancochrane (commented):
Oh, I also forgot to mention -- there are some merge conflicts with the main branch, do you mind resolving those?

db/raw/chicago_parks.geojson
wrangtangle: Great point. It's from https://data.cityofchicago.org/Parks-Recreation/Parks-Chicago-Park-District-Park-Boundaries-curren/ej32-qgdr. Added a comment in the Makefile noting this.

app/mbm/static/js/turnbyturn/displayDirections.js
wrangtangle: Agreed that this would be notionally better to do in CSS, but I'd suggest not bothering changing it in the interest of shipping. It's a lot of logic but not computationally intensive and shouldn't cause performance problems, so from the user perspective it should be the same. Let me know what you think.

db/label-parks.sql
wrangtangle: Great points! But `I'm also curious to hear if you think this whole thing is worth spending more time on at all` <- I'd say no? It feels like our time is better spent elsewhere for now. If analytics show people are requesting directions in/through parks, then that's a reason to work more on this.

app/mbm/directions.py
wrangtangle: Nope, good catch. I've removed it.

app/mbm/static/js/turnbyturn/displayDirections.js
jeancochrane: We'll test this out in staging and if it causes problems we can try switching to CSS. But this is probably good enough for now.

db/label-parks.sql
jeancochrane: We agreed that the 80/20 solution is just to preserve parentheticals for now, to make sure we don't accidentally remove important parts of park names. We can easily refine this later if if bothers us.

PR #74, opened 2025-11-16
Add major streets (20% or more of route length) to route info summary

wrangtangle:
For up to 3 streets that each make up 20% or more of the length of a route, add their names to the route info summary.

## Before

image

## After

image


wrangtangle:
Have resolved the merge conflicts!

PR #89, opened 2026-04-08
Use materialized view to filter out non-essential sidewalks

jeancochrane:
## Background

This PR proposes a new system for filtering out non-essential sidewalks when routing.

The motivating problem is that the number of ways in `chicago_ways` has more than tripled since 2020, from ~300k ways to ~1m ways. The vast majority (~500k) of these new ways are sidewalks ("footpaths" in OSM terminology). This is a major problem because `pgr_dijkstra` has [worse than linear time complexity](https://docs.pgrouting.org/3.1/en/pgr_dijkstra.html#description) (`O(V log V)`), so query times have more than tripled. In my experience, it's made routing unusable, though that is ultimately a subjective determination. Even if we were to down-weight sidewalks when routing (as in #70), we still face the problem of increased query times due to all the new sidewalks.

## Proposed solution

This PR combines two approaches to bring our query performance back in line:

1. **Create a new `clean_ways` materialized view that includes only essential sidewalks**. "Essential sidewalks" here describe sidewalks that are necessary to traverse in order to connect two tagged routes (off-street paths or mellow streets only). Essential sidewalks are recorded as a new `MellowRoute.type` and tagged from the `chicago_ways` table like all the other route types. Then, a database trigger refreshes the `clean_ways` view any time the `MellowRoute` table changes.
    - The **benefits** of this approach:
        1. We get to keep `chicago_ways` intact, and continue using it as the basis for tagging. This is useful since I expect we will want to tweak our set of tagged sidewalks in the future, which would be difficult if we full-on dropped them from the `chicago_ways` table.
        2. Minimal changes to the routing code itself, since `clean_ways` can act as a drop-in replacement for `chicago_ways`.
        3. Decently easy to extend if we want to, say, further reduce the base table for the routing query to exclude disconnected components (As in #86).
    - The **downsides** of this approach:
        1. Django can't manage materialized views out of the box. [There's a plugin](https://pypi.org/project/django-materialized-view/), but I don't want to introduce a new dependency unless we absolutely need to. So the materialized view is slightly less ergonomic to use than other models, particularly insofar as we need to write manual migrations. But since it's pretty simple, I don't think this outweighs the benefits of the view.
        2. Increased complexity, in that we need to refresh `clean_ways` any time we make edits to our mellow routes or to the underlying OSM data. We refresh automatically on changes to our mellow routes using a database trigger, but that also introduces complexity.
        3. Probably not the most intuitive data model design, so perhaps a long-term maintenance burden.
5. **Exclude all ways from the routing query if they are more than 2 miles outside the bounding box formed by the source and target nodes**. Even after filtering out sidewalks, I noticed that query performance had degraded due to the addition of the ~200k non-sidewalk ways. We had chatted about bounding our queries during a previous in-person discussion, so I decided to give it a try to see how it works. From my anecdotal experience, it does indeed improve query performance, dramatically so for sources and targets that are close to one another. Two miles seems like a generous buffer, but ultimately it's somewhat arbitrary, and I'm open to alternative suggestions.
    - **Benefits** of this approach:
        1. Faster query speeds, particularly for nearby sources/targets.
        2. Relatively simple to implement -- just requires a CTE and one additional filter in the routing query.  
    - **Downsides** of this approach:
        1. It has worked fine for every ad-hoc query I've QCed, but I still don't really know if, in aggregate, it will cause the query to return some non-intuitive routes, or cause a bunch of routes to fail to be computable at all. We can at least test for the increase in non-computable routes, as I'll describe in the [Testing](#testing) section below, but I can't think of a good way of testing for non-intuitive routes other than a lot of manual QC.

## Alternatives considered

I think the main alternative here might be to combine the bounding box query with down-weighting sidewalks along the lines of https://github.com/jeancochrane/mellow-bike-map/pull/86. My suspicion is that an approach like this would work equally well for short routes, but would cause worse and worse performance as the distance between the source and the target increases, to a scale where it would be as bad as no bounding box when e.g. traversing the city. As such, I went with the approach that drops non-essential sidewalks, because it felt like the maximally performant option in all cases. However, I'm interested in testing to see exactly how much more performant my proposed design is, because combining the bounding box with down-weighted sidewalks would be a dramatically less complex solution.

## Testing

I would like to do some fuzz testing to get a quantitative sense of how query times and routing distances compare between this diff, prod, and the alternate approach described in "Alternatives considered" above. I also want to check to see if the number of routes that cannot be found increases due to the new query bounding box. However, I wanted to get this PR up for design discussion ASAP, since it represents a pretty major refactor, and we might want to handle this problem differently. If the design looks good, I'll move ahead with fuzz testing as a QC step before the prod deploy.


app/mbm/templates/mbm/index.html
```diff
@@ -10,67 +10,70 @@
...
-              <i
-                data-toggle="tooltip"
-                data-placement="top"
-                title="V2 routing uses a new experimental algorithm that considers all residential streets to be mellow streets."
-                style="cursor: pointer"
-              >
-                <i class="fa fa-fw fa-info-circle text-muted"></i>
-              </i>
+  <!-- Hack to enable full-bleed map, since containers add margins by default -->
+  <div style="overflow-x: hidden">
```
jeancochrane, line 14: This diff fixes an unrelated bug that has been bothering me: The slight X-axis overflow beyond the right-hand edge of the map. Particularly on desktop, I was finding myself often scrolling to the right and breaking the full bleed of the map while doing intensive QC on routes:

Image

Aside from adding a top level `overflow-x: hidden` div, the rest of the changes to this file are caused by the added indentation level, so you can remove them by ignoring whitespace changes.

app/mbm/templates/mbm/mellow_route_list.html
```diff
@@ -76,6 +77,12 @@ <h3 class="py-3">
                     Edit paths
                   </a>
                 </td>
+                <td>
+                  <a href="{% url 'mellow-route-edit' object.slug view.model.Type.SIDEWALK %}">
+                    <i class="fa fa-fw fa-edit"></i>
+                    Edit sidewalks
+                  </a>
+                </td>
```
jeancochrane, line 85: This change adds sidewalks to the admin list view for tagging routes.

app/mbm/constants.py
```diff
@@ -0,0 +1,19 @@
...
+    101,  # cycleway:track
+    201,  # cycleway:right:track
+    301,  # cycleway:left:track
+    501,  # highway:cycleway
+)
+
+SIDEWALK_TAG_IDS = (
+    503,  # highway:pedestrian
+    504,  # highway:footway
+)
```
jeancochrane, line 19: I took a second to factor these out while I was here, since `SIDEWALK_TAG_IDS` is shared with a migration, so it can't be stored in `models.py`.

app/mbm/models.py
```diff
@@ -89,24 +90,39 @@ class Meta:
         unique_together = ('slug', 'type')
 
     @classmethod
-    def all(cls):
+    def all(cls, exclude_sidewalks=False):
```
jeancochrane, line 116: Many of the important sidewalks look weird on a map, since sidewalks ways tend to have larger geometries, and so they often don't match up exactly with the specific stretch that we need to connect two routes. As such, I wanted a way to exclude them from the main map that shows the user all of the mellow routes, to avoid confusion. This kwarg enables that change.

For an example of a weird sidewalk, see this sidewalk that is intended to connect the north/south route through Senn High School to the east/west route on Ardmore Ave -- we only really need the crossing, but tagging that crossing gives us the full south and southwest sidewalks:

Image

app/mbm/views.py
```diff
@@ -92,6 +77,9 @@ def get_nearest_vertex_id(self, coord):
             cursor.execute("""
                 SELECT vert.id
                 FROM chicago_ways_vertices_pgr AS vert
+                INNER JOIN clean_ways AS cw
+                    ON vert.id = cw.source
+                    OR vert.id = cw.target
```
jeancochrane, line 82: `chicago_ways_verticies_pgr` has vertices for all rows in `chicago_ways`, which includes sidewalks that we filter out of `clean_ways`. This inner join removes vertices for any ways that are not in `clean_ways`, and doesn't seem to incur any performance hit from my anecdotal QC.

app/mbm/views.py
```diff
@@ -126,7 +130,9 @@ def get_route(self, source_vertex_id, target_vertex_id, enable_v2=False):
                             {f"WHEN way.tag_id in {CYCLEWAY_TAG_IDS} THEN way.cost * 0.1" if enable_v2 is True else ""}
                             WHEN mellow.type = ''street'' THEN way.cost * 0.25
                             {f"WHEN way.tag_id in {RESIDENTIAL_STREET_TAG_IDS} THEN way.cost * 0.25" if enable_v2 is True else ""}
+                            WHEN mellow.type = ''sidewalk'' THEN way.cost * 0.3
                             WHEN way.oneway = ''YES'' THEN way.cost * 0.5
+                            WHEN way.tag_id in {RESIDENTIAL_STREET_TAG_IDS} THEN way.cost * 0.5
```
jeancochrane, line 135: This is a small quality-of-life improvement that I added in while QCing routes using `clean_ways`, but I could be convinced to pull it out into its own PR if it's controversial. Basically, I think all residential streets should be higher priority than streets tagged with the `route` type, or non-residential streets. For example, I was noticing the algorithm routing onto Kedzie for a few blocks to get to a destination rather than taking an obvious side street, just because Kedzie is tagged as a `route` and the side street wasn't tagged at all.

There's a bigger question regarding whether we want to prioritize `route` streets at all. I think it might be worth stripping them out, since I don't really care about busy streets with bike lanes. But that feels like a bigger change than this one, which just slightly upweights residential streets.

app/mbm/views.py
```diff
@@ -126,7 +130,9 @@ def get_route(self, source_vertex_id, target_vertex_id, enable_v2=False):
                             {f"WHEN way.tag_id in {CYCLEWAY_TAG_IDS} THEN way.cost * 0.1" if enable_v2 is True else ""}
                             WHEN mellow.type = ''street'' THEN way.cost * 0.25
                             {f"WHEN way.tag_id in {RESIDENTIAL_STREET_TAG_IDS} THEN way.cost * 0.25" if enable_v2 is True else ""}
+                            WHEN mellow.type = ''sidewalk'' THEN way.cost * 0.3
```
jeancochrane, line 133: 0.3 is somewhat arbitrary here: Higher priority than, say, a one-way, but lower priority than a mellow street. The idea is that we should only route onto sidewalks if they are connecting two tagged mellow streets or off-street paths. In my experience, this weighting works when combined with the fact that there are very few sidewalks, but it's possible there are edge cases I haven't tested yet.

scripts/after_install.sh
```diff
@@ -14,8 +14,10 @@ VENV_DIR="/home/mbm/.virtualenvs/$DEPLOYMENT_NAME"
 # the deployment specific folder
 mv /home/mbm/mellow-bike-map $PROJECT_DIR
 
-# Create a deployment specific virtual environment
-python3 -m venv $VENV_DIR
+# Create a deployment specific virtual environment.
+# Run as the mbm user in a login shell so that we pick up that user's configured
+# pyenv version, if one exists
+sudo -H -i -u mbm python3 -m venv $VENV_DIR
```
jeancochrane, line 20: We need to use pyenv to install Python on the staging server now, since Python 3.8 is so old. We should definitely upgrade it, but I'm punting that upgrade to a later date -- my goal here was to only upgrade packages that were absolutely necessary to support the change, which turned out to be the OS, postgres, and Jinja.

scripts/app_start.sh
```diff
@@ -63,7 +63,8 @@ for deployment in $old_deployments; do
     if [[ ! $deployment == $DEPLOYMENT_ID ]]; then
         echo "Signalling application processes from $deployment"
 
-        STATUS=`supervisorctl status $APP_NAME-$deployment:*`
+        STATUS=`supervisorctl status $APP_NAME-$deployment:* 2>/dev/null || echo 'NOT FOUND'`
+        echo "Supervisor status for $APP_NAME-$deployment: $STATUS"
```
jeancochrane, line 67: Handling an interesting edge case I had never run into before: If a deployment fails after it writes files, but before it is able to start up a new supervisor process, then there won't be any supervisor process whose status can be checked. That raises an error unless we handle it explicitly.

jeancochrane:
PR up here @wrangtangle! I'm curious to get your design feedback. After writing up all those comments, I'm realizing that the quantitative data is pretty important here, so I'm going to prioritize getting that all measured ASAP.

wrangtangle:
Extremely cool stuff! Excited to be able to get the latest OSM data onto the site.

Persisting my thoughts from yesterday:

- It occurs to me that it might wind up being faster to ship this if we split it up into two PRs. Though constraining the search space to the enlarged bounding box and the materialized view are both intended to mitigate the same problem of increased query times, they are independent improvements, and it might easier to focus on one, get it out, then focus on the other. Up to you of course.
- Regarding the bounding box:
  - Rather than pick two miles as a buffer arbitrarily, we could figure out a safe buffer programmatically. If we run a script that picks an arbitrary vertex pair, routes between them, finds the furthest vertex on the route outside the bounding box of the start/end vertices, then repeats this thousands of times and finds the max distance of a furthest vertex across all those runs, we can add a bit to that and then have a pretty safe buffer.
    - We could also run a version of this script that picks vertex pairs with very similar lat or lon values, since that would be the situation least tolerant of a winding route.
    - The buffer doesn't necessarily need to be a static value, but it could potentially be formulaically derived from the straight-line distance of the start/end points. If that works out, it might make routing between nearby points even faster.
  - We can always include named bike paths in the search space, since those are the most relevant candidates for winding far away from the bounding box.
  - If search within the bounding box fails to generate a route, we can fall back to searching outside of the bounding box.
- Regarding the `clean_ways` materialized view:
  - Since the "essential sidewalks" are manually tagged, we run the risk of missing some, as well as needing to manually retag as the OSM graph changes over time. Could we augment the manually tagged "essential sidewalks" with programmatically discovered ones? We brainstormed some different ideas for this, and I think the one that seems most promising to me, I forget whose it was, would go something like: pick two vertices, generate a route between them with sidewalks heavily downweighted, see if it contains any sidewalks, then for each contiguous section of sidewalk in the route, try banning it from being routed on and then regenerating the route. If the route gets markedly worse (as measured by length/score) then consider the ways in that contiguous segment of sidewalks "essential sidewalks". A process like that could be repeated many times to find remaining essential sidewalks.
  - If the query against `clean_ways` fails to find a route, we can fall back to querying all ways.
  - Bikeshedding, but the name `clean_ways` seems less informative than it could be. I think what we want to convey is "the ways that will be searched for routing queries (potentially with a fallback to a superset)." I'd suggest `routable_ways`, though this isn't perfect, since it implies that other ways are not capable of being routed on. Another possibility that is maybe more accurate but less clear is `tier_one_ways`, if we allow a fallback to a superset.

app/mbm/views.py
wrangtangle: I think the need for a judgment call based on personal preference here is evidence that we should eventually let users supply their own weights (in whatever sort of friendly UX makes that easiest).

jeancochrane: Yup, agreed!

jeancochrane:
Thanks for the detailed thoughts, @wrangtangle! I did some quick and dirty benchmarking today along the lines we discussed on Monday:

<details>
<summary>Expand for benchmarking script code</summary>

```python
#!/usr/bin/env python3
"""
Benchmark three routing approaches for mellow-bike-map (PR #89).

Approach A: clean_ways materialized view + 2-mile buffered bounding box
Approach B: chicago_ways + same bbox (no clean_ways filtering)
Approach C: chicago_ways, no bbox (main branch baseline)

Usage:
    # From inside the Docker app container:
    python /app/scripts/benchmark_routing.py

    # From the host (with Postgres exposed on port 32001):
    DATABASE_URL=postgresql://postgres:postgres@localhost:32001/mbm \\
        python scripts/benchmark_routing.py
"""

import math
import os
import statistics
import time

import psycopg2

# Assume the script is running in the app container for database access
DEFAULT_DATABASE_URL = "postgres://postgres:postgres@postgres:5432/mbm"
NUM_TRIALS = 5
NUM_WARMUP = 1

# pgRouting/osm2pgrouting tag IDs (matches mbm/constants.py)
RESIDENTIAL_TAG_IDS = (507, 509)

# Chicago test pairs: (source_lat, source_lng, target_lat, target_lng, label)
TEST_PAIRS = [
    (41.9098, -87.6775, 41.9148, -87.6710, "short-~0.5mi-wicker-park"),
    (41.8450, -87.6600, 41.8570, -87.6510, "short-~1mi-pilsen-to-bridgeport"),
    (41.9277, -87.7086, 41.9098, -87.6775, "medium-~2mi-logan-sq-to-wicker"),
    (41.7780, -87.6480, 41.8350, -87.6540, "medium-~4mi-englewood-to-bridgeport"),
    (41.9683, -87.6897, 41.8900, -87.6300, "long-~6mi-lincoln-sq-to-river-north"),
    (41.8350, -87.6540, 41.8950, -87.7653, "long-~8mi-bridgeport-to-austin"),
    (42.0075, -87.6684, 41.8559, -87.6511, "xlong-~10mi-rogers-park-to-pilsen"),
]


def connect():
    url = os.getenv("DATABASE_URL", DEFAULT_DATABASE_URL)
    return psycopg2.connect(url)


def find_vertex_a(cursor, lat, lng):
    """Nearest vertex that exists in clean_ways (approach A)."""
    cursor.execute(
        """
        SELECT vert.id
        FROM chicago_ways_vertices_pgr AS vert
        INNER JOIN clean_ways AS cw
            ON vert.id = cw.source OR vert.id = cw.target
        ORDER BY vert.the_geom <-> ST_SetSRID(ST_MakePoint(%s, %s), 4326)
        LIMIT 1
        """,
        [lng, lat],
    )
    row = cursor.fetchone()
    return row[0] if row else None


def find_vertex_bc(cursor, lat, lng):
    """Nearest vertex across all ways (approaches B & C)."""
    cursor.execute(
        """
        SELECT id
        FROM chicago_ways_vertices_pgr
        ORDER BY the_geom <-> ST_SetSRID(ST_MakePoint(%s, %s), 4326)
        LIMIT 1
        """,
        [lng, lat],
    )
    row = cursor.fetchone()
    return row[0] if row else None


def route_a(cursor, src, tgt):
    """Approach A: clean_ways + bbox."""
    cursor.execute(
        f"""
        SELECT way.gid, way.length_m
        FROM pgr_dijkstra(
            'WITH mellow AS (
                SELECT DISTINCT(UNNEST(ways)) AS osm_id, type
                FROM mbm_mellowroute
            ),
            bbox AS (
                SELECT ST_Expand(
                    ST_Envelope(ST_Collect(the_geom)), 0.029
                ) AS geom
                FROM chicago_ways_vertices_pgr
                WHERE id IN ({src}, {tgt})
            )
            SELECT
                way.gid AS id,
                way.source,
                way.target,
                CASE
                    WHEN mellow.type = ''''path'''' THEN way.cost * 0.1
                    WHEN mellow.type = ''''street'''' THEN way.cost * 0.25
                    WHEN mellow.type = ''''sidewalk'''' THEN way.cost * 0.3
                    WHEN way.oneway = ''''YES'''' THEN way.cost * 0.5
                    WHEN way.tag_id in {RESIDENTIAL_TAG_IDS}
                        THEN way.cost * 0.5
                    WHEN mellow.type = ''''route'''' THEN way.cost * 0.75
                    ELSE way.cost
                END AS cost,
                CASE
                    WHEN mellow.type = ''''path'''' THEN way.reverse_cost * 0.1
                    WHEN mellow.type = ''''street'''' THEN way.reverse_cost * 0.25
                    WHEN mellow.type = ''''sidewalk'''' THEN way.reverse_cost * 0.3
                    WHEN way.oneway = ''''YES'''' THEN way.reverse_cost * 0.5
                    WHEN way.tag_id in {RESIDENTIAL_TAG_IDS}
                        THEN way.reverse_cost * 0.5
                    WHEN mellow.type = ''''route'''' THEN way.reverse_cost * 0.75
                    ELSE way.reverse_cost
                END AS reverse_cost
            FROM clean_ways AS way
            JOIN bbox ON way.the_geom && bbox.geom
            LEFT JOIN mellow USING(osm_id)',
            {src},
            {tgt}
        ) AS path
        JOIN clean_ways AS way ON path.edge = way.gid
        """
    )
    return cursor.fetchall()


def route_b(cursor, src, tgt):
    """Approach B: chicago_ways + bbox (no clean_ways filtering)."""
    cursor.execute(
        f"""
        SELECT way.gid, way.length_m
        FROM pgr_dijkstra(
            'WITH mellow AS (
                SELECT DISTINCT(UNNEST(ways)) AS osm_id, type
                FROM mbm_mellowroute
            ),
            bbox AS (
                SELECT ST_Expand(
                    ST_Envelope(ST_Collect(the_geom)), 0.029
                ) AS geom
                FROM chicago_ways_vertices_pgr
                WHERE id IN ({src}, {tgt})
            )
            SELECT
                way.gid AS id,
                way.source,
                way.target,
                CASE
                    WHEN mellow.type = ''''path'''' THEN way.cost * 0.1
                    WHEN mellow.type = ''''street'''' THEN way.cost * 0.25
                    WHEN way.oneway = ''''YES'''' THEN way.cost * 0.5
                    WHEN mellow.type = ''''route'''' THEN way.cost * 0.75
                    ELSE way.cost
                END AS cost,
                CASE
                    WHEN mellow.type = ''''path'''' THEN way.reverse_cost * 0.1
                    WHEN mellow.type = ''''street'''' THEN way.reverse_cost * 0.25
                    WHEN way.oneway = ''''YES'''' THEN way.reverse_cost * 0.5
                    WHEN mellow.type = ''''route'''' THEN way.reverse_cost * 0.75
                    ELSE way.reverse_cost
                END AS reverse_cost
            FROM chicago_ways AS way
            JOIN bbox ON way.the_geom && bbox.geom
            LEFT JOIN mellow USING(osm_id)',
            {src},
            {tgt}
        ) AS path
        JOIN chicago_ways AS way ON path.edge = way.gid
        """
    )
    return cursor.fetchall()


def route_c(cursor, src, tgt):
    """Approach C: chicago_ways, no bbox (main branch baseline)."""
    cursor.execute(
        f"""
        SELECT way.gid, way.length_m
        FROM pgr_dijkstra(
            'WITH mellow AS (
                SELECT DISTINCT(UNNEST(ways)) AS osm_id, type
                FROM mbm_mellowroute
            )
            SELECT
                way.gid AS id,
                way.source,
                way.target,
                CASE
                    WHEN mellow.type = ''''path'''' THEN way.cost * 0.1
                    WHEN mellow.type = ''''street'''' THEN way.cost * 0.25
                    WHEN way.oneway = ''''YES'''' THEN way.cost * 0.5
                    WHEN mellow.type = ''''route'''' THEN way.cost * 0.75
                    ELSE way.cost
                END AS cost,
                CASE
                    WHEN mellow.type = ''''path'''' THEN way.reverse_cost * 0.1
                    WHEN mellow.type = ''''street'''' THEN way.reverse_cost * 0.25
                    WHEN way.oneway = ''''YES'''' THEN way.reverse_cost * 0.5
                    WHEN mellow.type = ''''route'''' THEN way.reverse_cost * 0.75
                    ELSE way.reverse_cost
                END AS reverse_cost
            FROM chicago_ways AS way
            LEFT JOIN mellow USING(osm_id)',
            {src},
            {tgt}
        ) AS path
        JOIN chicago_ways AS way ON path.edge = way.gid
        """
    )
    return cursor.fetchall()


def time_route(cursor, route_fn, src, tgt, trials, warmup):
    """Run route_fn(src, tgt) warmup+trials times; return (times_ms, routable)."""
    times = []
    routable = None

    for i in range(warmup + trials):
        t0 = time.perf_counter()
        rows = route_fn(cursor, src, tgt)
        elapsed_ms = (time.perf_counter() - t0) * 1000

        if i < warmup:
            continue  # discard warm-up

        routable = len(rows) > 0
        times.append(elapsed_ms)

    return times, routable


def haversine_mi(lat1, lng1, lat2, lng2):
    """Straight-line distance in miles between two lat/lng points."""
    R = 3958.8  # Earth radius in miles
    phi1, phi2 = math.radians(lat1), math.radians(lat2)
    dphi = math.radians(lat2 - lat1)
    dlam = math.radians(lng2 - lng1)
    a = math.sin(dphi / 2) ** 2 + math.cos(phi1) * math.cos(phi2) * math.sin(dlam / 2) ** 2
    return R * 2 * math.atan2(math.sqrt(a), math.sqrt(1 - a))


def fmt(times):
    if not times:
        return "n/a"
    med = statistics.median(times)
    mn = statistics.mean(times)
    sd = statistics.stdev(times) if len(times) > 1 else 0
    return f"{med:6.0f}ms (mean {mn:.0f}, sd {sd:.0f})"


def main():
    conn = connect()
    cursor = conn.cursor()

    approaches = [
        ("A (clean_ways+bbox)", route_a),
        ("B (chicago_ways+bbox)", route_b),
        ("C (main baseline)", route_c),
    ]

    results = {}  # (pair_label, approach_label) -> (times, routable)

    for src_lat, src_lng, tgt_lat, tgt_lng, label in TEST_PAIRS:
        dist_mi = haversine_mi(src_lat, src_lng, tgt_lat, tgt_lng)
        print(f"\n{'='*70}")
        print(f"Pair: {label}  ({dist_mi:.1f} mi straight-line)")
        print(f"  Source: ({src_lat}, {src_lng})  Target: ({tgt_lat}, {tgt_lng})")
        print(f"{'='*70}")

        src_a = find_vertex_a(cursor, src_lat, src_lng)
        tgt_a = find_vertex_a(cursor, tgt_lat, tgt_lng)
        src_bc = find_vertex_bc(cursor, src_lat, src_lng)
        tgt_bc = find_vertex_bc(cursor, tgt_lat, tgt_lng)

        print(f"  Vertices — A: {src_a}→{tgt_a}  B/C: {src_bc}→{tgt_bc}")

        for approach_label, route_fn in approaches:
            if approach_label.startswith("A"):
                src_id, tgt_id = src_a, tgt_a
            else:
                src_id, tgt_id = src_bc, tgt_bc

            if src_id is None or tgt_id is None:
                print(f"  {approach_label}: SKIP (vertex not found)")
                results[(label, approach_label)] = ([], False)
                continue

            times, routable = time_route(
                cursor, route_fn, src_id, tgt_id, NUM_TRIALS, NUM_WARMUP
            )
            results[(label, approach_label)] = (times, routable)
            status = "✓ routed" if routable else "✗ no route"
            print(f"  {approach_label:24s}: {fmt(times)}  [{status}]")

    # Summary comparison table
    print(f"\n{'='*70}")
    print("SUMMARY — median execution time (ms)")
    print(f"{'='*70}")
    col_w = 26
    header = f"{'Pair':<28}" + "".join(f"{a[0]:>{col_w}}" for a, _ in approaches)
    print(header)
    print("-" * len(header))
    for src_lat, src_lng, tgt_lat, tgt_lng, label in TEST_PAIRS:
        row = f"{label:<28}"
        for approach_label, _ in approaches:
            times, routable = results.get((label, approach_label), ([], False))
            if times and routable:
                cell = f"{statistics.median(times):.0f}ms"
            elif not routable:
                cell = "no route"
            else:
                cell = "n/a"
            row += f"{cell:>{col_w}}"
        print(row)

    cursor.close()
    conn.close()


if __name__ == "__main__":
    main()
```

</details>

The results strike me as quite promising for just the bounding box approach without tagged sidewalks:

<details>
<summary>Expand for benchmarking script results</summary>

```
======================================================================
Pair: short-~0.5mi-wicker-park  (0.5 mi straight-line)
  Source: (41.9098, -87.6775)  Target: (41.9148, -87.671)
======================================================================
  Vertices — A: 127393→217091  B/C: 538990→217091
  A (clean_ways+bbox)     :    117ms (mean 116, sd 3)  [✓ routed]
  B (chicago_ways+bbox)   :    205ms (mean 213, sd 22)  [✓ routed]
  C (main baseline)       :   4948ms (mean 4887, sd 138)  [✓ routed]

======================================================================
Pair: short-~1mi-pilsen-to-bridgeport  (0.9 mi straight-line)
  Source: (41.845, -87.66)  Target: (41.857, -87.651)
======================================================================
  Vertices — A: 284013→22215  B/C: 284013→22215
  A (clean_ways+bbox)     :    116ms (mean 114, sd 7)  [✓ routed]
  B (chicago_ways+bbox)   :    271ms (mean 271, sd 9)  [✓ routed]
  C (main baseline)       :   4820ms (mean 4799, sd 79)  [✓ routed]

======================================================================
Pair: medium-~2mi-logan-sq-to-wicker  (2.0 mi straight-line)
  Source: (41.9277, -87.7086)  Target: (41.9098, -87.6775)
======================================================================
  Vertices — A: 35239→127393  B/C: 35239→538990
  A (clean_ways+bbox)     :    172ms (mean 181, sd 15)  [✓ routed]
  B (chicago_ways+bbox)   :    405ms (mean 405, sd 14)  [✓ routed]
  C (main baseline)       :   4667ms (mean 4686, sd 196)  [✓ routed]

======================================================================
Pair: medium-~4mi-englewood-to-bridgeport  (4.0 mi straight-line)
  Source: (41.778, -87.648)  Target: (41.835, -87.654)
======================================================================
  Vertices — A: 229208→215052  B/C: 229208→215052
  A (clean_ways+bbox)     :    141ms (mean 150, sd 15)  [✓ routed]
  B (chicago_ways+bbox)   :    381ms (mean 388, sd 23)  [✓ routed]
  C (main baseline)       :   4893ms (mean 4858, sd 221)  [✓ routed]

======================================================================
Pair: long-~6mi-lincoln-sq-to-river-north  (6.2 mi straight-line)
  Source: (41.9683, -87.6897)  Target: (41.89, -87.63)
======================================================================
  Vertices — A: 144640→414373  B/C: 144640→423246
  A (clean_ways+bbox)     :    349ms (mean 356, sd 22)  [✓ routed]
  B (chicago_ways+bbox)   :    958ms (mean 963, sd 28)  [✓ routed]
  C (main baseline)       :   4922ms (mean 4923, sd 200)  [✓ routed]

======================================================================
Pair: long-~8mi-bridgeport-to-austin  (7.1 mi straight-line)
  Source: (41.835, -87.654)  Target: (41.895, -87.7653)
======================================================================
  Vertices — A: 215052→316795  B/C: 215052→497000
  A (clean_ways+bbox)     :    455ms (mean 452, sd 11)  [✓ routed]
  B (chicago_ways+bbox)   :   1205ms (mean 1201, sd 21)  [✓ routed]
  C (main baseline)       :   5320ms (mean 5496, sd 335)  [✓ routed]

======================================================================
Pair: xlong-~10mi-rogers-park-to-pilsen  (10.5 mi straight-line)
  Source: (42.0075, -87.6684)  Target: (41.8559, -87.6511)
======================================================================
  Vertices — A: 180862→75452  B/C: 180862→75452
  A (clean_ways+bbox)     :    352ms (mean 355, sd 11)  [✓ routed]
  B (chicago_ways+bbox)   :    967ms (mean 973, sd 13)  [✓ routed]
  C (main baseline)       :   5068ms (mean 5006, sd 118)  [✓ routed]
```

</details>

If these preliminary results scale, then the bounding box is the main performance improvement that this PR offers, and the tagged sidewalks are only an incremental improvement on top of that. Plus, I think you're right that we should tune the bounding box buffer (and add a fallback in case it fails), and a smaller buffer could offer an additional incremental improvement.

Per your suggestion, I'm going to extract out the bounding box into a separate PR and do some tuning on the buffer in that context.

jeancochrane:
Draft PR with just the bounding box change (and a few of the small bug fixes + tweaks to support the new staging site) up here: https://github.com/jeancochrane/mellow-bike-map/pull/90 I'll plan to post further fuzz testing and buffer tuning results in that PR.

wrangtangle:
Super interesting to see the benchmarking results!

It would also have been interesting to how `clean_ways` without the bounding box did, for a one-to-one comparison with the results for `(chicago_ways+bbox)`.

Even 117ms, with both improvements, feels slow for a 0.5-mile straight-line route search. I think the +2 mile on each side buffer for the bounding box is probably way larger than necessary for 0.5 miles, so it will be interesting to see how performance changes if we conclude we're able to safely reduce the buffer.

jeancochrane:
> It would also have been interesting to how clean_ways without the bounding box did, for a one-to-one comparison with the results for (chicago_ways+bbox).

I thought this was a good point, so I updated the script to test it! Here's a new version of the script with slightly better abstractions, to make it easier to configure a matrix of source table + bounding box size:

<details>
<summary>Expand for updated benchmarking script code</summary>

```python
#!/usr/bin/env python3
#!/usr/bin/env python3
"""
Benchmark three routing approaches for mellow-bike-map (PR #89).

Approach A: clean_ways materialized view
Approach B: clean_ways + 2-mile buffered bounding box
Approach C: chicago_ways (main branch baseline)
Approach D: chicago_ways + same bbox (no clean_ways filtering)

Usage:
    # From inside the Docker app container:
    python /app/scripts/benchmark_routing.py

    # From the host (with Postgres exposed on port 32001):
    DATABASE_URL=postgresql://postgres:postgres@localhost:32001/mbm \\
        python scripts/benchmark_routing.py
"""

import math
import os
import statistics
import time

import psycopg2

# Assume the script is running in the app container for database access
DEFAULT_DATABASE_URL = "postgres://postgres:postgres@postgres:5432/mbm"
NUM_TRIALS = 5
NUM_WARMUP = 1

# pgRouting/osm2pgrouting tag IDs (matches mbm/constants.py)
RESIDENTIAL_TAG_IDS = (507, 509)

# Chicago test pairs: (source_lat, source_lng, target_lat, target_lng, label)
TEST_PAIRS = [
    (41.9098, -87.6775, 41.9148, -87.6710, "short-~0.5mi-wicker-park"),
    (41.8450, -87.6600, 41.8570, -87.6510, "short-~1mi-pilsen-to-bridgeport"),
    (41.9277, -87.7086, 41.9098, -87.6775, "medium-~2mi-logan-sq-to-wicker"),
    (41.7780, -87.6480, 41.8350, -87.6540, "medium-~4mi-englewood-to-bridgeport"),
    (41.9683, -87.6897, 41.8900, -87.6300, "long-~6mi-lincoln-sq-to-river-north"),
    (41.8350, -87.6540, 41.8950, -87.7653, "long-~8mi-bridgeport-to-austin"),
    (42.0075, -87.6684, 41.8559, -87.6511, "xlong-~10mi-rogers-park-to-pilsen"),
]


def connect():
    url = os.getenv("DATABASE_URL", DEFAULT_DATABASE_URL)
    return psycopg2.connect(url)


def miles_to_degrees(miles, latitude = 41.88):
    """
    Convert a distance in miles to decimal degrees suitable for use with
    PostGIS ST_Expand() / ST_Envelope() on geometries in SRID 4326 (WGS 84).

    Because one degree of longitude shrinks with latitude, this function
    returns the *larger* of the two degree values (latitude degrees) so that
    the resulting bounding box is at least `miles` wide in every direction.

    Args:
        miles: Distance in miles to expand the bounding box.
        latitude: Reference latitude in decimal degrees. Defaults to Chicago's
                  approximate center (~41.88°N).

    Returns:
        A degree value to pass to ST_Expand().

    Example PostGIS usage:
        SELECT ST_Envelope(
            ST_Expand(
                ST_SetSRID(ST_MakePoint(lng, lat), 4326),
                miles_to_degrees(5)
            )
        );
    """
    # 1 degree of latitude ≈ 69.0 miles (constant everywhere)
    lat_degrees_per_mile = 1 / 69.0

    # 1 degree of longitude ≈ 69.0 * cos(latitude) miles
    lng_degrees_per_mile = 1 / (69.0 * math.cos(math.radians(latitude)))

    # Use the larger value so the box covers `miles` in all directions
    return miles * max(lat_degrees_per_mile, lng_degrees_per_mile)


def find_vertex(cursor, lat, lng, ways_tbl):
    """Nearest vertex that exists in clean_ways."""
    cursor.execute(
        f"""
        SELECT vert.id
        FROM chicago_ways_vertices_pgr AS vert
        {
            f'''
             INNER JOIN clean_ways AS cw
                 ON vert.id = cw.source OR vert.id = cw.target   
            '''
            if ways_tbl == 'clean_ways'
            else ''
        }
        ORDER BY vert.the_geom <-> ST_SetSRID(ST_MakePoint(%s, %s), 4326)
        LIMIT 1
        """,
        [lng, lat],
    )
    row = cursor.fetchone()
    return row[0] if row else None


def route(cursor, src_id, tgt_id, ways_tbl, bbox_mi = None):
    """Route between nodes with gids `src_id` and `tgt_id`, using `ways_tbl`
    as the source table for the ways. If `bbox_mi` is not null, restrict ways
    in the search space to a bounding box around the source and target node,
    and give the box a buffer corresponding to `bbox_mi` miles."""
    cursor.execute(
        f"""
        SELECT way.gid, way.length_m
        FROM pgr_dijkstra(
            'WITH mellow AS (
                SELECT DISTINCT(UNNEST(ways)) AS osm_id, type
                FROM mbm_mellowroute
            )
            {
                f'''
                ,
                bbox AS (
                    SELECT ST_Expand(
                        ST_Envelope(ST_Collect(the_geom)),
                        {miles_to_degrees(bbox_mi)}
                    ) AS geom
                    FROM chicago_ways_vertices_pgr
                    WHERE id IN ({src_id}, {tgt_id})
                )
                '''
                if bbox_mi is not None
                else ""
            }
            SELECT
                way.gid AS id,
                way.source,
                way.target,
                CASE
                    WHEN mellow.type = ''path'' THEN way.cost * 0.1
                    WHEN mellow.type = ''street'' THEN way.cost * 0.25
                    WHEN way.oneway = ''YES'' THEN way.cost * 0.5
                    WHEN mellow.type = ''route'' THEN way.cost * 0.75
                    ELSE way.cost
                END AS cost,
                CASE
                    WHEN mellow.type = ''path'' THEN way.reverse_cost * 0.1
                    WHEN mellow.type = ''street'' THEN way.reverse_cost * 0.25
                    WHEN way.oneway = ''YES'' THEN way.reverse_cost * 0.5
                    WHEN mellow.type = ''route'' THEN way.reverse_cost * 0.75
                    ELSE way.reverse_cost
                END AS reverse_cost
            FROM {ways_tbl} AS way
            {'JOIN bbox ON way.the_geom && bbox.geom' if bbox_mi is not None else ""}
            LEFT JOIN mellow USING(osm_id)',
            {src_id},
            {tgt_id}
        ) AS path
        JOIN {ways_tbl} AS way ON path.edge = way.gid
        """
    )
    return cursor.fetchall()


def time_route(cursor, src, tgt, trials, warmup, ways_tbl, bbox_mi):
    """Run route_fn(cursor, src, tgt, ways_tbl, bbox_mi) warmup+trials times;
    return (times_ms, routable)."""
    times = []
    routable = None

    for i in range(warmup + trials):
        t0 = time.perf_counter()
        rows = route(cursor, src, tgt, ways_tbl, bbox_mi)
        elapsed_ms = (time.perf_counter() - t0) * 1000

        if i < warmup:
            continue  # discard warm-up

        routable = len(rows) > 0
        times.append(elapsed_ms)

    return times, routable


def haversine_mi(lat1, lng1, lat2, lng2):
    """Straight-line distance in miles between two lat/lng points."""
    R = 3958.8  # Earth radius in miles
    phi1, phi2 = math.radians(lat1), math.radians(lat2)
    dphi = math.radians(lat2 - lat1)
    dlam = math.radians(lng2 - lng1)
    a = math.sin(dphi / 2) ** 2 + math.cos(phi1) * math.cos(phi2) * math.sin(dlam / 2) ** 2
    return R * 2 * math.atan2(math.sqrt(a), math.sqrt(1 - a))


def fmt(times):
    if not times:
        return "n/a"
    med = statistics.median(times)
    mn = statistics.mean(times)
    sd = statistics.stdev(times) if len(times) > 1 else 0
    return f"{med:6.0f}ms (mean {mn:.0f}, sd {sd:.0f})"


def main():
    conn = connect()
    cursor = conn.cursor()

    approaches = [
        ("clean_ways", "clean_ways", None),
        ("clean_ways+0.5mi-bbox", "clean_ways", 0.5),
        ("clean_ways+1mi-bbox", "clean_ways", 1),
        ("clean_ways+2mi-bbox", "clean_ways", 2),
        ("chicago_ways", "chicago_ways", None),
        ("chicago_ways+0.5mi-bbox", "chicago_ways", 0.5),
        ("chicago_ways+1mi-bbox", "chicago_ways", 1),
        ("chicago_ways+2mi-bbox", "chicago_ways", 2),
    ]

    results = {}

    for src_lat, src_lng, tgt_lat, tgt_lng, label in TEST_PAIRS:
        dist_mi = haversine_mi(src_lat, src_lng, tgt_lat, tgt_lng)
        print(f"\n{'='*70}")
        print(f"Pair: {label}  ({dist_mi:.1f} mi straight-line)")
        print(f"  Source: ({src_lat}, {src_lng})  Target: ({tgt_lat}, {tgt_lng})")

        src_clean_ways = find_vertex(cursor, src_lat, src_lng, "clean_ways")
        tgt_clean_ways = find_vertex(cursor, tgt_lat, tgt_lng, "clean_ways")
        src_chicago_ways = find_vertex(cursor, src_lat, src_lng, "chicago_ways")
        tgt_chicago_ways = find_vertex(cursor, tgt_lat, tgt_lng, "chicago_ways")
        print(
            f"  Vertices — clean_ways {src_clean_ways}→{tgt_clean_ways}, "
            f"chicago_ways {src_chicago_ways}→{tgt_chicago_ways}"
        )

        print(f"{'='*70}")

        for approach_label, ways_tbl, bbox_mi in approaches:
            src_id = src_clean_ways if ways_tbl == "clean_ways" else src_chicago_ways
            tgt_id = tgt_clean_ways if ways_tbl == "clean_ways" else tgt_chicago_ways
            if src_id is None or tgt_id is None:
                print(f"  {approach_label}: SKIP (vertex not found)")
                results[(label, approach_label)] = ([], False)
                continue

            times, routable = time_route(
                cursor, src_id, tgt_id, NUM_TRIALS, NUM_WARMUP, ways_tbl, bbox_mi
            )
            results[(label, approach_label)] = (times, routable)
            status = "✓ routed" if routable else "✗ no route"
            print(f"  {approach_label:24s}: {fmt(times)}  [{status}]")

    # Summary comparison table
    print(f"\n{'='*70}")
    print("SUMMARY — median execution time (ms)")
    print(f"{'='*70}")
    col_w = 26
    header = f"{'Pair':<28}" + "".join(f"{a[0]:>{col_w}}" for a, _ in approaches)
    print(header)
    print("-" * len(header))
    for src_lat, src_lng, tgt_lat, tgt_lng, label in TEST_PAIRS:
        row = f"{label:<28}"
        for approach_label, _ in approaches:
            times, routable = results.get((label, approach_label), ([], False))
            if times and routable:
                cell = f"{statistics.median(times):.0f}ms"
            elif not routable:
                cell = "no route"
            else:
                cell = "n/a"
            row += f"{cell:>{col_w}}"
        print(row)

    cursor.close()
    conn.close()


if __name__ == "__main__":
    main()
```

</details>

<details>
<summary>Expand for detailed results</summary>

```
======================================================================
Pair: short-~0.5mi-wicker-park  (0.5 mi straight-line)
  Source: (41.9098, -87.6775)  Target: (41.9148, -87.671)
  Vertices — clean_ways 127393→217091, chicago_ways 538990→217091
======================================================================
  clean_ways              :   2194ms (mean 2202, sd 77)  [✓ routed]
  clean_ways+0.5mi-bbox   :     24ms (mean 25, sd 3)  [✓ routed]
  clean_ways+1mi-bbox     :     50ms (mean 51, sd 1)  [✓ routed]
  clean_ways+2mi-bbox     :    182ms (mean 181, sd 5)  [✓ routed]
  chicago_ways            :   5085ms (mean 5113, sd 55)  [✓ routed]
  chicago_ways+0.5mi-bbox :     36ms (mean 37, sd 4)  [✓ routed]
  chicago_ways+1mi-bbox   :    117ms (mean 115, sd 4)  [✓ routed]
  chicago_ways+2mi-bbox   :    455ms (mean 456, sd 4)  [✓ routed]

======================================================================
Pair: short-~1mi-pilsen-to-bridgeport  (0.9 mi straight-line)
  Source: (41.845, -87.66)  Target: (41.857, -87.651)
  Vertices — clean_ways 284013→22215, chicago_ways 284013→22215
======================================================================
  clean_ways              :   2135ms (mean 2102, sd 74)  [✓ routed]
  clean_ways+0.5mi-bbox   :     26ms (mean 26, sd 1)  [✓ routed]
  clean_ways+1mi-bbox     :     55ms (mean 54, sd 1)  [✓ routed]
  clean_ways+2mi-bbox     :    181ms (mean 180, sd 5)  [✓ routed]
  chicago_ways            :   5134ms (mean 5120, sd 27)  [✓ routed]
  chicago_ways+0.5mi-bbox :     47ms (mean 47, sd 2)  [✓ routed]
  chicago_ways+1mi-bbox   :    135ms (mean 135, sd 5)  [✓ routed]
  chicago_ways+2mi-bbox   :    529ms (mean 535, sd 10)  [✓ routed]

======================================================================
Pair: medium-~2mi-logan-sq-to-wicker  (2.0 mi straight-line)
  Source: (41.9277, -87.7086)  Target: (41.9098, -87.6775)
  Vertices — clean_ways 35239→127393, chicago_ways 35239→538990
======================================================================
  clean_ways              :   2167ms (mean 2147, sd 36)  [✓ routed]
  clean_ways+0.5mi-bbox   :     49ms (mean 50, sd 1)  [✓ routed]
  clean_ways+1mi-bbox     :    103ms (mean 103, sd 2)  [✓ routed]
  clean_ways+2mi-bbox     :    281ms (mean 280, sd 5)  [✓ routed]
  chicago_ways            :   5097ms (mean 5116, sd 62)  [✓ routed]
  chicago_ways+0.5mi-bbox :    118ms (mean 117, sd 4)  [✓ routed]
  chicago_ways+1mi-bbox   :    250ms (mean 252, sd 8)  [✓ routed]
  chicago_ways+2mi-bbox   :    660ms (mean 653, sd 12)  [✓ routed]

======================================================================
Pair: medium-~4mi-englewood-to-bridgeport  (4.0 mi straight-line)
  Source: (41.778, -87.648)  Target: (41.835, -87.654)
  Vertices — clean_ways 229208→215052, chicago_ways 229208→215052
======================================================================
  clean_ways              :   2004ms (mean 2008, sd 21)  [✓ routed]
  clean_ways+0.5mi-bbox   :     39ms (mean 39, sd 1)  [✓ routed]
  clean_ways+1mi-bbox     :     82ms (mean 82, sd 2)  [✓ routed]
  clean_ways+2mi-bbox     :    223ms (mean 226, sd 6)  [✓ routed]
  chicago_ways            :   5084ms (mean 5141, sd 86)  [✓ routed]
  chicago_ways+0.5mi-bbox :     95ms (mean 95, sd 1)  [✓ routed]
  chicago_ways+1mi-bbox   :    199ms (mean 201, sd 4)  [✓ routed]
  chicago_ways+2mi-bbox   :    650ms (mean 648, sd 20)  [✓ routed]

======================================================================
Pair: long-~6mi-lincoln-sq-to-river-north  (6.2 mi straight-line)
  Source: (41.9683, -87.6897)  Target: (41.89, -87.63)
  Vertices — clean_ways 144640→414373, chicago_ways 144640→423246
======================================================================
  clean_ways              :   2186ms (mean 2179, sd 67)  [✓ routed]
  clean_ways+0.5mi-bbox   :    185ms (mean 188, sd 7)  [✓ routed]
  clean_ways+1mi-bbox     :    251ms (mean 250, sd 8)  [✓ routed]
  clean_ways+2mi-bbox     :    436ms (mean 437, sd 10)  [✓ routed]
  chicago_ways            :   5126ms (mean 5084, sd 148)  [✓ routed]
  chicago_ways+0.5mi-bbox :    489ms (mean 498, sd 21)  [✓ routed]
  chicago_ways+1mi-bbox   :    738ms (mean 737, sd 28)  [✓ routed]
  chicago_ways+2mi-bbox   :   1243ms (mean 1254, sd 30)  [✓ routed]

======================================================================
Pair: long-~8mi-bridgeport-to-austin  (7.1 mi straight-line)
  Source: (41.835, -87.654)  Target: (41.895, -87.7653)
  Vertices — clean_ways 215052→316795, chicago_ways 215052→497000
======================================================================
  clean_ways              :   2357ms (mean 2459, sd 151)  [✓ routed]
  clean_ways+0.5mi-bbox   :    253ms (mean 246, sd 21)  [✓ routed]
  clean_ways+1mi-bbox     :    356ms (mean 357, sd 9)  [✓ routed]
  clean_ways+2mi-bbox     :    607ms (mean 608, sd 11)  [✓ routed]
  chicago_ways            :   5480ms (mean 5657, sd 276)  [✓ routed]
  chicago_ways+0.5mi-bbox :    529ms (mean 531, sd 12)  [✓ routed]
  chicago_ways+1mi-bbox   :    957ms (mean 958, sd 11)  [✓ routed]
  chicago_ways+2mi-bbox   :   1683ms (mean 1686, sd 18)  [✓ routed]

======================================================================
Pair: xlong-~10mi-rogers-park-to-pilsen  (10.5 mi straight-line)
  Source: (42.0075, -87.6684)  Target: (41.8559, -87.6511)
  Vertices — clean_ways 180862→75452, chicago_ways 180862→75452
======================================================================
  clean_ways              :   2865ms (mean 2826, sd 305)  [✓ routed]
  clean_ways+0.5mi-bbox   :    188ms (mean 190, sd 23)  [✓ routed]
  clean_ways+1mi-bbox     :    265ms (mean 276, sd 28)  [✓ routed]
  clean_ways+2mi-bbox     :    510ms (mean 525, sd 32)  [✓ routed]
  chicago_ways            :   5887ms (mean 5938, sd 299)  [✓ routed]
  chicago_ways+0.5mi-bbox :    402ms (mean 399, sd 15)  [✓ routed]
  chicago_ways+1mi-bbox   :    701ms (mean 714, sd 51)  [✓ routed]
  chicago_ways+2mi-bbox   :   1421ms (mean 1449, sd 138)  [✓ routed]
```

</details>

TL;DR:`clean_ways` is a little more than twice as fast as `chicago_ways`, which makes sense because `pgr_dijkstra()` is slightly worse than linear and `clean_ways` contains roughly half as many ways as `chicago_ways`. The 2-mile buffered bounding box introduces a speedup somewhere between 75-95%; reducing the bounding box size introduces further improvement that is roughly proportional to the reduction in size. Put together, these two results suggest that `chicago_ways` with a 1-mile bounding box is similar to `clean_ways` with a 2-mile bounding box.

To me, these results confirm that the bounding box is the right incremental step, and tuning the bounding box offers potentially large performance gains. I'm going to move onto that tuning step, and I'll log my results in #90.

README.md
```diff
@@ -14,18 +14,23 @@ Build containers:
...
 ```
 
+Import the data (note that the full data import can take quite a bit of memory,
+so make sure to adjust your Docker preferences to allow Docker at least 6GB of
+RAM):
+
+```
+docker compose run --rm -w /app postgres make db/import/chicago.table
+```
+
```
kalilsn, line 24: This should still come after the migrations, right? And should this run in the app container rather than the postgres container? Right now it's failing for me since the Makefile is expecting to find the database at the host `postgres` (rather than localhost).

kalilsn: Ok nevermind, I guess the intent is just to have already started the database container with `docker compose up -d` or similar before running this command?

app/mbm/views.py
kalilsn: I don't really think this part is controversial, but I am curious what you're thinking about doing with the v2 routing feature. This looks like one half of the v2 algorithm. What do you think about eliminating the v2 option entirely and merging the other part of it into the regular routing query? To me, lowering the cost of the tagged protected/off-street paths is even less controversial than prioritizing residential streets!

app/mbm/views.py
```diff
@@ -116,6 +110,16 @@ def get_route(self, source_vertex_id, target_vertex_id, enable_v2=False):
                     'WITH mellow AS (
                         SELECT DISTINCT(UNNEST(ways)) AS osm_id, type
                         FROM mbm_mellowroute
+                    ),
+                    -- Restrict input ways to a 2-mile bounding box around the
+                    -- source and target to limit the complexity of the query
+                    bbox AS (
+                        SELECT ST_Expand(
+                            ST_Envelope(ST_Collect(the_geom)),
+                            0.029  -- 2 miles
```
kalilsn, line 119: This is close enough to 2 miles latitude in chicago, but the conversion between miles and degrees longitude varies quite a bit at different latitudes. You could fix this by using the variant of ST_Expand that accepts different distances for x and y expansion. Alternatively, [this answer](https://gis.stackexchange.com/a/298696) suggests that you could cast to geography and use `_ST_Expand` which would default to using miles as the unit.

README.md
jeancochrane: I think that's correct, though I will update the instructions to make it clear that we expect users to bring up the database container before loading data or running migrations!

app/mbm/views.py
jeancochrane: Good point! I'm extracting this logic out into #90 (haven't gotten around to updating this PR to reflect that change yet) so I mirrored your comment over there: https://github.com/jeancochrane/mellow-bike-map/pull/90/changes#r3126720051

jeancochrane:
I think #90 provided a more elegant solution for the same problem, so I'm closing this PR but leaving the branch up. Performance is currently acceptable by my standards, but we can revisit this approach if we decide we need even better performance.

PR #90, opened 2026-04-16
Use a buffered bounding box to restrict ways in routing query

jeancochrane:
## Background

This PR pulls out the most significant performance improvement from https://github.com/jeancochrane/mellow-bike-map/pull/89 into its own PR: Using a buffered bounding box between the source and target coordinates to restrict the ways that `pgr_dijkstra()` sees when choosing a route.

Based on testing, this should provide at least a 2.5x improvement in query speed over the main branch. The change is up on staging and performance feels roughly comparable to prod to me.

## Details

The PR folds together a number of changes as part of this feature:

- **Adds a dynamic buffer to the bounding box**, whose size is the lesser of either A) 1/2 the straight-line distance between the source and the target, or B) 2 miles
    - This buffer formula is somewhat arbitrary, but it felt good to me during testing; 1/2 is easy to understand as a heuristic for reducing the buffer for short routes, and 2 miles feels to me like the maximum distance I would go out of my way for a chiller route based on some testing with my most common routes
- **Excludes off-street paths from the bounding box**, so that the routing algorithm always includes them; this allows a route to continue on an off-street path like the Lakefront Trail even if it enters and exits the bounding box temporarily
- Adds **fallback behavior** so that if a route cannot be found between the source and target inside the bounding box, the backend will send another query to try to find a route _without_ the bounding box
- Adds an **optional `showBbox` URL param** that controls whether the app will display the bounding box on the map when it returns a route; this is useful for debugging the bounding box
- Tweaks the function that returns the nearest vertex to a point (`get_nearest_vertex_id()`) so that it **won't return sidewalks**; this is helpful since otherwise routes tend to always start and end on sidewalks, since a sidewalk is almost always the closest way to a coordinate that represents a building
- Tweaks the routing query to give **higher priority to residential streets**. The new order of priority from highest to lowest:
    1. Off-street paths (mellow tag `path`)
    2. Chill streets (mellow tag `street`)
    3. One-way streets and residential streets (OSM tags)
    4. Main streets with bike lanes (mellow tag `route`)
    5. All other way types
 - Tweaks the database setup commands in the Makefile to **add an index on `chicago_ways(osm_id)`**, which provides roughly a 5x speedup on staging compared to the routing query without the index


wrangtangle:
As other potential TODOs, are you thinking about always allowing named bike paths in the search space, and falling back to searching outside the bounding box if searching within it fails?

jeancochrane:
Yes, thanks for the reminder @wrangtangle! I just updated the TODO section in the PR body.

app/mbm/views.py
```diff
@@ -116,6 +131,16 @@ def get_route(self, source_vertex_id, target_vertex_id, enable_v2=False):
                     'WITH mellow AS (
                         SELECT DISTINCT(UNNEST(ways)) AS osm_id, type
                         FROM mbm_mellowroute
+                    ),
+                    -- Restrict input ways to a 2-mile bounding box around the
+                    -- source and target to limit the complexity of the query
+                    bbox AS (
+                        SELECT ST_Expand(
+                            ST_Envelope(ST_Collect(the_geom)),
+                            0.029  -- 2 miles
```
jeancochrane: TODO: We probably need to make this more robust to differences in degree size across latitudes. See: https://github.com/jeancochrane/mellow-bike-map/pull/89#discussion_r3126628688

jeancochrane: Done in [c8fa361](https://github.com/jeancochrane/mellow-bike-map/pull/90/commits/c8fa3619d451318999e4a0d17a0e5534ae6544f4).

app/mbm/views.py
```diff
@@ -104,71 +121,61 @@ def get_nearest_vertex_id(self, coord):
...
-                JOIN chicago_ways AS way
-                ON path.edge = way.gid
-                LEFT JOIN (
-                    SELECT DISTINCT(UNNEST(ways)) AS osm_id, type
-                    FROM mbm_mellowroute
-                ) as mellow
-                USING(osm_id)
-            """, [source_vertex_id, target_vertex_id])
-            rows = fetchall(cursor)
+    def get_route(
```
jeancochrane, line 124: I've refactored this function pretty extensively to make it easier to test. Unfortunately, that makes the diff harder to read. But the basic gist is that I've extracted out a handful of functions:

- `_build_route_query()`: Takes the input parameters and returns a string representing the SQL query that can generate a route; useful for testing conditional logic, since most arguments to this function just serve to alter the contents of the SQL query
- `_execute_route_query()`: Thin wrapper around `_build_route_query()` that executes the SQL query and returns results; useful in order to implement the fallback logic that fires in case no route can be found in the bounding box, in which case we need to resend the query with the bounding box removed
- `_build_bbox_query()`: Generate SQL string representing bounding box query
- `_execute_bbox_query()`: Thin wrapper around `_build_bbox_query()` that executes the SQL and returns a geoJSON feature

These functions are listed in the module in rough order from top to bottom of the call stack.

app/mbm/views.py
```diff
@@ -182,6 +192,260 @@ def get_route(self, source_vertex_id, target_vertex_id, enable_v2=False):
...
+                        way.target,
+                        way.cost,
+                        way.reverse_cost,
+                        way.oneway,
+                        way.tag_id,
+                        mellow.type
+                    FROM chicago_ways AS way
+                    JOIN mellow USING(osm_id)
+                    WHERE mellow.type = ''path''
+                        AND NOT (way.the_geom && (SELECT geom FROM bbox))
```
jeancochrane: I tried a number of different approaches to this query, and this was the one that performed the best. Will edit this comment to document the other approaches and their timings (on my laptop) below later today.

app/mbm/views.py
```diff
@@ -182,6 +192,260 @@ def get_route(self, source_vertex_id, target_vertex_id, enable_v2=False):
...
+                    id,
+                    source,
+                    target,
+                    CASE
+                        WHEN type = ''path'' THEN cost * 0.1
+                        {f"WHEN tag_id in {CYCLEWAY_TAG_IDS} THEN cost * 0.1" if enable_v2 is True else ""}
+                        WHEN type = ''street'' THEN cost * 0.25
+                        {f"WHEN tag_id in {RESIDENTIAL_STREET_TAG_IDS} THEN cost * 0.25" if enable_v2 is True else ""}
+                        WHEN oneway = ''YES'' THEN cost * 0.5
+                        WHEN tag_id in {RESIDENTIAL_STREET_TAG_IDS} THEN cost * 0.5
```
kalilsn: This will end up lowering the cost of residential streets during v2 queries even further. Not very important since these numbers are a bit arbitrary but it might make sense to move this into the conditional on 324 to avoid that

app/mbm/views.py
```diff
@@ -182,6 +192,260 @@ def get_route(self, source_vertex_id, target_vertex_id, enable_v2=False):
...
+                        WHEN type = ''route'' THEN cost * 0.75
+                        ELSE cost
+                    END AS cost,
+                    CASE
+                        WHEN type = ''path'' THEN reverse_cost * 0.1
+                        {f"WHEN tag_id in {CYCLEWAY_TAG_IDS} THEN cost * 0.1" if enable_v2 is True else ""}
+                        WHEN type = ''street'' THEN reverse_cost * 0.25
+                        {f"WHEN tag_id in {RESIDENTIAL_STREET_TAG_IDS} THEN cost * 0.25" if enable_v2 is True else ""}
+                        WHEN oneway = ''YES'' THEN reverse_cost * 0.5
+                        WHEN tag_id in {RESIDENTIAL_STREET_TAG_IDS} THEN cost * 0.5
```
kalilsn: This isn't new, but should these `cost`s be `reverse_cost`?

jeancochrane: Thank you so much for catching this! Fixed in fb860fbf67904d1ae9d3198a2b39a9459bb9dacc.

app/mbm/views.py
jeancochrane: I'm not totally sure I understand. Here's my mental model of how a non-tagged residential street would pass through this pattern match:

- If `enable_v2 is True`, the following case will trigger, decreasing the cost of the residential street by 4x (`cost * 0.25`):

```python
{f"WHEN tag_id in {RESIDENTIAL_STREET_TAG_IDS} THEN cost * 0.25" if enable_v2 is True else ""}
```

- Otherwise, the following case will trigger, decreasing the cost of the residential street by 2x (`cost * 0.5`):

```python
WHEN tag_id in {RESIDENTIAL_STREET_TAG_IDS} THEN cost * 0.5
```

So when `enable_v2` is active, we'll continue assigning the same 4x priority to residential streets as before this PR, but when `enable_v2` is not active, residential streets will get a smaller 2x priority --  a higher priority than they get on the main branch, but less than V2. Is that also how you read it?

I definitely think the V2 priority doesn't make a whole lot of sense, but I'm going to rip that out separately in #101, since I want to remove it from the frontend too.

app/mbm/views.py
jeancochrane: I actually tweaked this yet again -- follow-up comment to come on the most recent version of the code.

app/mbm/views.py
```diff
@@ -182,6 +192,261 @@ def get_route(self, source_vertex_id, target_vertex_id, enable_v2=False):
...
+                        way.target,
+                        way.cost,
+                        way.reverse_cost,
+                        way.oneway,
+                        way.tag_id,
+                        mellow.type
+                    FROM mellow
+                    JOIN chicago_ways AS way USING(osm_id)
+                    WHERE mellow.type = 'path'
+                )
```
jeancochrane, line 286: Two other options I considered for this `edge` CTE:

1. Removing the `UNION` and just joining `chicago_ways` to `mellow`, then filtering `WHERE mellow.type = 'path' OR way.the_geom && (SELECT geom FROM bbox)`. This is slightly slower than the `UNION` approach because Postgres has to do a full table scan on `chicago_ways` to check for `mellow.type = 'path'` and so doesn't get to make use of the spatial index on `chicago_ways.the_geom` when intersecting with the bounding box.

```sql
 SELECT
    way.gid AS id,
     way.source,
     way.target,
     way.cost,
     way.reverse_cost,
     way.oneway,
     way.tag_id,
     mellow.type
 FROM chicago_ways AS way
 LEFT JOIN mellow USING(osm_id)
 WHERE mellow.type = 'path'
     OR way.the_geom && (SELECT geom FROM bbox)
```

2. Using `UNION ALL` and finding paths in the symmetric difference of `chicago_ways` and the bbox. I thought that maybe this would be helpful because Postgres could skip the deduplication step involved in a `UNION` but in practice it doesn't seem to make much of a difference, I think because the number of ways with `mellow.type = 'path'` is so small.

```sql
    SELECT
        way.gid AS id,
        way.source,
        way.target,
        way.cost,
        way.reverse_cost,
        way.oneway,
        way.tag_id,
        mellow.type
    FROM chicago_ways AS way
    JOIN bbox ON way.the_geom && bbox.geom
    LEFT JOIN mellow USING(osm_id)
    UNION ALL
    SELECT
        way.gid AS id,
        way.source,
        way.target,
        way.cost,
        way.reverse_cost,
        way.oneway,
        way.tag_id,
        mellow.type
    FROM mellow
    JOIN chicago_ways AS way USING(osm_id)
    WHERE mellow.type = 'path'
        AND NOT (way.the_geom && (SELECT geom FROM bbox))
```

I feel like I might be missing something obvious so let me know if you have different ideas for this query! I bet there's more low-hanging fruit we could pick for performance.

app/tests/app.start.test.mjs
```diff
@@ -314,3 +314,61 @@ test('manual coordinate search removes existing address params', () => {
   assert.equal(params.get('targetCoordinates'), '42.1,-87.8')
   assert.equal(params.get('utm_source'), 'duckduckgo.com')
 })
+
+test('search sends show_bbox=true to backend when showBbox URL param is present', async () => {
```
jeancochrane: Full disclosure: I don't really understand how our JavaScript test suite works so I did have an LLM generate these tests for me. I think it's worth giving them a little extra scrutiny.

app/mbm/views.py
jeancochrane: Thinking about this more, I'm realizing that this change isn't really related to the main goal of the PR, and it might be more appropriate to save it for #101 anyway! Let me know if you feel strongly about it one way or the other.

jeancochrane:
I think this is ready for review @kalilsn! I made one final tweak to simplify the `edge` CTE, left comments on the trickiest bits, fixed one bug you caught, and deployed the latest version of the code to staging.

app/mbm/views.py
kalilsn: You are completely right! My bad! And I support you pulling it out anyways.

kalilsn (approved):

app/mbm/views.py
jeancochrane: Removed in c9ff87f0fddafb5897ce7b7d364381978727cd22.

jeancochrane:
Calling out that I removed the frontend tests in bfc289c2a171df772a979d7d3eb939cb8c0ad94e following an in-person conversation where we agreed that they aren't adding much value.

PR #91, opened 2026-04-22
Small tweaks to support the new staging EC2 instance

jeancochrane:
This PR makes a couple small changes to improve deployments for the EC2 staging server we set up in https://github.com/jeancochrane/mellow-bike-map/pull/88. See comments below for details.

### DIFF
--- a/configs/staging-config.conf
+++ b/configs/staging-config.conf
@@ -1,2 +1,2 @@
 APP_NAME=mellow-bike-map
-export DOMAIN=staging-ec2.mellowbikemap.com
+export DOMAIN=staging.mellowbikemap.com
--- a/scripts/after_install.sh
+++ b/scripts/after_install.sh
@@ -14,8 +14,10 @@ VENV_DIR="/home/mbm/.virtualenvs/$DEPLOYMENT_NAME"
 # the deployment specific folder
 mv /home/mbm/mellow-bike-map $PROJECT_DIR
 
-# Create a deployment specific virtual environment
-python3 -m venv $VENV_DIR
+# Create a deployment specific virtual environment.
+# Run as the mbm user in a login shell so that we pick up that user's configured
+# pyenv version, if one exists
+sudo -H -i -u mbm python3 -m venv $VENV_DIR
 
 # Set the ownership of the project files and the virtual environment
 chown -R mbm.www-data $PROJECT_DIR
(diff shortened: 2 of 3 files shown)


configs/staging-config.conf
```diff
@@ -1,2 +1,2 @@
 APP_NAME=mellow-bike-map
-export DOMAIN=staging-ec2.mellowbikemap.com
+export DOMAIN=staging.mellowbikemap.com
```
jeancochrane, line 2: This change lets the EC2 staging server use the official `staging` subdomain. I'm doing this now because I'm confident that EC2 is the right deployment platform for staging, at least unless/until we decide to migrate to another infra provider.

scripts/after_install.sh
```diff
@@ -14,8 +14,10 @@ VENV_DIR="/home/mbm/.virtualenvs/$DEPLOYMENT_NAME"
 # the deployment specific folder
 mv /home/mbm/mellow-bike-map $PROJECT_DIR
 
-# Create a deployment specific virtual environment
-python3 -m venv $VENV_DIR
+# Create a deployment specific virtual environment.
+# Run as the mbm user in a login shell so that we pick up that user's configured
+# pyenv version, if one exists
+sudo -H -i -u mbm python3 -m venv $VENV_DIR
```
jeancochrane, line 20: The app runs a version of Python that is no longer available in apt, so we need to install it with a Python version management tool (like pyenv). I don't want to couple the deployment scripts to the assumption that we're using pyenv, so instead we just create the venv dir as the `mbm` user so that we can use whatever Python version is configured as the default for that user. This mostly matches the pattern we use for `pip` commands on e.g. lines 30-31 anyway, it's just slightly different in that we're using a login shell to make sure the bash profile gets loaded to enable shims.

scripts/after_install.sh
```diff
@@ -68,7 +70,7 @@ fi
 
 # Install Jinja into the virtual environment and run the render_configs.py
 # script.
-$VENV_DIR/bin/pip install Jinja2==2.10
+$VENV_DIR/bin/pip install "Jinja2>=2.10,<3.2"
```
jeancochrane, line 73: Annoyingly, the prod server is on an older version of Python (3.6) than the staging server (3.8). Hopefully soon we will promote the staging server to prod and resolve this discrepancy, but in the meantime, this version range is necessary to support both Python versions.

scripts/app_start.sh
```diff
@@ -63,7 +63,8 @@ for deployment in $old_deployments; do
     if [[ ! $deployment == $DEPLOYMENT_ID ]]; then
         echo "Signalling application processes from $deployment"
 
-        STATUS=`supervisorctl status $APP_NAME-$deployment:*`
+        STATUS=`supervisorctl status $APP_NAME-$deployment:* 2>/dev/null || echo 'NOT FOUND'`
+        echo "Supervisor status for $APP_NAME-$deployment: $STATUS"
```
jeancochrane, line 67: This change was inspired by a weird edge case I had to debug: If a deployment fails after it has written supervisor configs, but before it has started that supervisor job, then you can end up with a weird state that raises an error when you try to run `supervisorctl status`. This change handles that edge case gracefully, and falls back to the generic status "NOT FOUND" if it occurs.

PR #103, opened 2026-05-11
Remove V2 routing toggle and improve priority order in the routing algorithm

### LINKED ISSUE #101: Merge v2 features into main routing query
- Drop conditionals from routing query
- Remove checkbox from UI
- Remove any v2 option api handling code

jeancochrane:
This PR makes two related changes:

1. Folds the most consequential logic from V2 routing into the main routing query, so as to render V2 unnecessary
2. Strips out the V2 routing option (i.e. the frontend checkbox and `enable_v2` kwarg to routing methods)

The routing algorithm now follows this system of weights when determining the cost of a way:

1. Off-street bike paths that we have tagged with a `MellowRoute` instance using the `"path"` type (10% weight)
2. Streets that we have tagged with a `MellowRoute` instance using the `"street"` type (25% weight)
3. All other residential streets, protected bike paths, or off-street bike paths (50% weight)
4. All other one-way streets (75% weight)
5. All other types of ways that are not covered by one or more category above

This change is deployed to the staging site for testing.

Closes #101.


kalilsn (approved):
This looks great and is very exciting to me! I tested this a bit and the routes feel a lot better (meaning they more closely match where I usually ride)! I did encounter a few errors but I think they're just related to issues with the geolocation features.

Issue #31, opened 2021-08-21
Change site branding


jeancochrane:
Switch to new domain with new site name.

PR #104, opened 2026-05-12
Set up logging and error monitoring through sentry

### LINKED ISSUE #32: Add analytics
Prefer a self-hosted solution.

### LINKED ISSUE #96: Wire up logging and alerting
The app is already configured for Sentry, we just have to wire it up.

- Upgrade Sentry to the latest version
- Figure out how to update environment variables (get rid of Blackbox?)
- Maybe add Sentry to the frontend, if it's easy
- Possibly tracing, if it's free and easy

kalilsn:
Fixes #96
Fixes #32 

This PR sets up our app to send errors and logs from the front and back end to Sentry. In addition, I've added a django middleware that outputs json logs on every request, which we'll use to replace our analytics service.

A few notes:
- To set the SENTRY_DSN env var I had to remove Bea's key from the blackbox admins and create the encrypted conf file on one of our deployed serves. We should definitely test this out on staging first.
- We're loading sentry via the loader script since we don't have a package manager and bundler set up


app/mbm/middleware.py
```diff
@@ -0,0 +1,32 @@
...
+        start_time = time.time()
+
+        response: HttpResponse = self.get_response(request)
+
+        duration = time.time() - start_time
+        extra = {"duration": duration, "method": request.method,
+                 "path": request.path, "status": response.status_code}
+
+        logged_query_params = ["source", "target", "showBbox", "sourceAddress",
+                               "targetAddress", "sourceCoordinates", "targetCoordinates"]
```
kalilsn, line 23: After I "fixed" the showBbox param I realized it was just following a pattern we've been using for all of the params. Leaving it as is for now but I'm happy to adjust however you like.

kalilsn: Returning to this after seeing how the sentry logs interface actually works, I actually think it would be good to get this list as small as possible for easier aggregation, since you have to select each column individually. I sent a few logs from my local machine and saved [a simple query that you're welcome to modify](https://chill-streets.sentry.io/explore/logs/?aggregateField=%7B%22yAxes%22%3A%5B%22count%28message%29%22%5D%7D&id=1434507&interval=5m&logsFields=timestamp&logsFields=message&logsFields=method&logsFields=path&logsFields=tags%5Bduration%2Cnumber%5D&logsFields=timestamp&logsFields=showBbox&logsFields=sourceAddress&logsFields=targetAddress&logsFields=source&logsFields=target&logsQuery=&logsSortBys=-timestamp&mode=samples&project=4511287443914752&statsPeriod=24h&title=requests)

app/mbm/logging.py
```diff
@@ -9,3 +12,10 @@ def before_send(event, hint):
         if log_record.name == 'django.security.DisallowedHost':
             event['fingerprint'] = ['disallowed-host']
     return event
+
+# For future logging customizations
+
+
+class JSONFormatter(json_log_formatter.JSONFormatter):
+    def json_record(self, message, extra, record):
+        return super().json_record(message, extra, record)
```
jeancochrane: **[Nitpick, optional]** Moving the comment to a class docstring to avoid drift in the future, since it's currently not totally clear the comment is intended to apply to the class definition:

```suggestion
class JSONFormatter(json_log_formatter.JSONFormatter):
    """Custom JSON log formatter. Currently passes through directly to the
	parent class, but we define a custom class anyway to make it easy to extend
	functionality in the future."""

    def json_record(self, message, extra, record):
        return super().json_record(message, extra, record)
```

app/mbm/templates/mbm/base.html
```diff
@@ -96,6 +80,46 @@
...
     <!-- Load external scripts -->
+    {% if enable_sentry %}
+    <script
+      src="https://js.sentry-cdn.com/611811e9bba3c86de0da65b84f82727e.min.js"
+      crossorigin="anonymous"
+    ></script>
+    <script>
+      window.sentryOnLoad = function () {
+        Sentry.init({
+          // Adds request headers and IP for users, for more info visit:
```
jeancochrane: **[Nitpick, optional]** I assume these are default comments included in the sample code, but I think we should tweak them so that they reflect the customizations we've made:


```suggestion
          // Don't log sensitive user info like IP addresses. For more info visit:
```

app/mbm/templates/mbm/base.html
```diff
@@ -96,6 +80,46 @@
...
+          // https://docs.sentry.io/platforms/javascript/configuration/options/#sendDefaultPii
+          sendDefaultPii: false,
+          // Alternatively, use `process.env.npm_package_version` for a dynamic release version
+          // if your build tool supports it.
```
jeancochrane: **[Nitpick, optional]** This doesn't apply to us:

```suggestion
```

app/mbm/templates/mbm/base.html
```diff
@@ -96,6 +80,46 @@
...
+          release: "{{ sentry_release }}",
+          integrations: [
+            Sentry.browserTracingIntegration(),
+            Sentry.replayIntegration(),
+          ],
+          // Enable logs to be sent to Sentry
```
jeancochrane: **[Nitpick, optional]** Tweaking to reflect the ground truth:


```suggestion
          // Don't send logs to Sentry. Our frontend logs are not yet
		  // structured enough for it to be useful to capture them
```

app/mbm/templates/mbm/base.html
```diff
@@ -96,6 +80,46 @@
...
+          enableLogs: false,
+          // Set tracesSampleRate to 1.0 to capture 100%
+          // of transactions for tracing.
+          // We recommend adjusting this value in production
+          // Learn more at
+          // https://docs.sentry.io/platforms/javascript/configuration/options/#traces-sample-rate
```
jeancochrane: **[Nitpick, optional]** Looks like the heading is actually slightly different on the page:


```suggestion
          // https://docs.sentry.io/platforms/javascript/configuration/options/#tracesSampleRate
```

app/mbm/templates/mbm/base.html
```diff
@@ -96,6 +80,46 @@
...
+          release: "{{ sentry_release }}",
+          integrations: [
+            Sentry.browserTracingIntegration(),
+            Sentry.replayIntegration(),
+          ],
+          // Enable logs to be sent to Sentry
+          enableLogs: false,
+          // Set tracesSampleRate to 1.0 to capture 100%
+          // of transactions for tracing.
+          // We recommend adjusting this value in production
```
jeancochrane, line 104: **[Question, non-blocking]** Do you have thoughts on whether 100% is the appropriate amount of samples to trace? It makes sense to me if we're sure Sentry will just stop collecting traces rather than start charging us if we surpass our limit. Otherwise, it might be prudent to reduce the number of traces we collect.

app/mbm/templates/mbm/base.html
```diff
@@ -96,6 +80,46 @@
...
+          // Learn more at
+          // https://docs.sentry.io/platforms/javascript/configuration/options/#traces-sample-rate
+          tracesSampleRate: 1.0,
+          // Set `tracePropagationTargets` to control for which URLs trace propagation should be enabled
+          tracePropagationTargets: [
+            /^https:\/\/(staging\.)?mellowbikemap\.com/,
```
jeancochrane, line 110: **[Suggestion, optional]** I wonder if it's worth defining this base hostname in a variable in `settings.py` to make it easier to update in the future? I could see easily see myself forgetting to update this value if we ever change the hostname in the future (e.g. in #31). On the other hand, maybe it's not a big deal, since I would expect us to at least do a quick code search for the old hostname before we change it. I don't feel strongly about it, so I leave it up to you to decide how important you think this is.

app/mbm/middleware.py
jeancochrane: **[Praise]** I think this list is a good start! I can imagine myself wanting to analyze the use of all of these params at some point, which indicates to me that we should keep them in.

app/mbm/middleware.py
```diff
@@ -0,0 +1,32 @@
...
+
+        for key in logged_query_params:
+            value = request.GET.get(key)
+            if value:
+                extra[key] = value
+
+        logger.info("response sent", extra=extra)
```
jeancochrane: **[Nitpick, optional]** I might be being overly defensive here, but I wonder if it's worth giving this log a more distinct message, since we plan to use it as the basis for analytics searches in the future? I could imagine the text "response sent" accidentally colliding with log messages from other services or even other parts of our own codebase. Here's one idea for a tweak to make it more unique:

```suggestion
        logger.info("mbm.response_sent", extra=extra)
```

app/mbm/settings.py
```diff
@@ -31,15 +31,49 @@
 allowed_hosts = os.getenv('DJANGO_ALLOWED_HOSTS', [])
 ALLOWED_HOSTS = allowed_hosts.split(',') if allowed_hosts else []
 
-
 # Configure Sentry for error logging
-if os.getenv('SENTRY_DSN'):
+SENTRY_RELEASE = os.path.basename(os.getcwd())
```
jeancochrane: **[Suggestion, optional]** This feels a bit brittle to me, since it relies on assumptions about the organization of our deployment directory and the working directory of our supervisor process, and we are likely to change those assumptions in the future without realizing our app is tightly coupled to them. This logic is also likely to fail silently in those cases (i.e. not raise an error but return a nonsensical value).

I totally forgot that we write a module `app/mbm/deployment.py` during deploys that we use to record the deployment ID:

https://github.com/jeancochrane/mellow-bike-map/blob/ddbc942089a64de611fd4a8ef0ebc978e19483b5/scripts/after_install.sh#L76-L78

Here's how we use it in the `pong()` endpoint to extract the deployment ID:

https://github.com/jeancochrane/mellow-bike-map/blob/ddbc942089a64de611fd4a8ef0ebc978e19483b5/app/mbm/views.py#L586-L594

I wonder if there's a way to repurpose this module for our purposes? One idea would be to move the import logic from `pong()` into `settings.py`, and then have `pong()` return `settings.DEPLOYMENT_ID`. The only wrinkle with that idea is that we need to figure out what to do in case of an import error: Do we raise the error if we're not in a dev environment, and thereby insist that all non-dev environments must have deployment IDs? Or do we suppress the error, fall back to `DEPLOYMENT_ID = None` and only raise an error in the `pong()` endpoint if `settings.DEPLOYMENT_ID` is null?

Personally, I lean towards moving the deployment ID parsing logic into `settings.py` and raising an error in non-dev environments if we encounter an `ImportError`, since it feels like the easiest system to reason about. But at the end of the day I'm happy with any of these options, including the current version of the code, if we decide neither of my suggested alternatives are simple enough to be worth pursuing. Either way, I'm hopeful that we'll switch to a saner deployment system in the near future, in which case we'll have to switch to a new value for `SENTRY_RELEASE` anyway.

jeancochrane (changes requested):
Awesome work! I'm really excited to have easily-accessible logs and stack traces to work with. No blocking concerns from my end, though I raised a number of small questions and suggestions below, so I'll wait to approve in order to prompt myself to take one final look at everything after you resolve my comments.

app/mbm/templates/mbm/base.html
kalilsn: We get 5M spans/month free and I haven't given them a card, so I think it's fine to capture 100% on our tiny app! But I don't really know that they'll be useful or anything so I don't care what we set this to.

app/mbm/templates/mbm/base.html
kalilsn: hilarious that they messed that up

app/mbm/templates/mbm/base.html
kalilsn: Probably a good idea 👍

app/mbm/middleware.py
kalilsn: I'm fine with this and also open to adding more info here to make debugging easier. For example, we could use our custom logging class to add the file/line number to every log. Let me know!

app/mbm/settings.py
kalilsn: I like your plan and I'll implement it! Though I'm not sure what happens if we accidentally deploy something that errors here. Is there some kind of rollback or deployment failure logic or will we just break prod?

app/mbm/middleware.py
jeancochrane: That's a fun idea! I'm fine with keeping it simple for now, but it's good to know we can build on it in the future.

app/mbm/settings.py
jeancochrane: Great question! In theory, the deployment scripts are designed to support blue-green deployments. The relevant logic is in [`scripts/app_start.sh`](https://github.com/jeancochrane/mellow-bike-map/blob/master/scripts/app_start.sh). First, we add the Supervisor config for the new app, keeping the old one in place, and start the new app running on its socket:

https://github.com/jeancochrane/mellow-bike-map/blob/95dbee7ac26342e72932c5d6c6a60e7c3611af4c/scripts/app_start.sh#L11-L13

Then, we ping the socket directly to make sure the app is healthy:

https://github.com/jeancochrane/mellow-bike-map/blob/95dbee7ac26342e72932c5d6c6a60e7c3611af4c/scripts/app_start.sh#L23-L36

Only once we've confirmed the app is healthy do we reload Nginx, thereby switching the proxy to point to the new socket:

https://github.com/jeancochrane/mellow-bike-map/blob/95dbee7ac26342e72932c5d6c6a60e7c3611af4c/scripts/app_start.sh#L47-L55

Here's the Nginx config, to demonstrate that the app points to a deployment-specific socket:

https://github.com/jeancochrane/mellow-bike-map/blob/95dbee7ac26342e72932c5d6c6a60e7c3611af4c/configs/mellow-bike-map-production.conf.nginx#L1-L7

So in theory, if the app fails to start, then the `app_start` script should error out at the healthcheck stage and Nginx should continue sending traffic to the prior deployment.

app/mbm/templates/mbm/base.html
kalilsn: Decided I don't want to do this that badly if you don't care. I'd really need to store the pattern `mellowbikemap\.com` which I don't think is actually going to help anything. Plus this is very unimportant config!

jeancochrane: Fine by me!

jeancochrane (approved):
Let's give it a try! Are the prod configs ready to go, such that when we merge this PR and kick off a new deployment, everything will just work?

PR #106, opened 2026-05-12
Check for presence of calm routes layer before setting style

### LINKED ISSUE #102: Race condition between calm routes layer and directions layer in `app.js`
I haven't debugged this super extensively, but I'm pretty sure there's a race condition between the calm routes layer and the directions layer in `app.js`. Particularly this line inside the `$.getJSON()` callback that displays the directions layer on the map:

https://github.com/jeancochrane/mellow-bike-map/blob/94ccbdf790f24e1d76ed4ea6a7b9bb747cca6c00/app/mbm/static/js/app.js#L593-L594

The calm routes layer is set by a separate async `$.getJSON()` callback that runs on startup:

https://github.com/jeancochrane/mellow-bike-map/blob/94ccbdf790f24e1d76ed4ea6a7b9bb747cca6c00/app/mbm/static/js/app.js#L295-L302

Since the directions layer callback can fire immediately on page load if source/target URL params are set, I believe it will trigger a race condition if it completes ahead of the calm routes layer:

```
! Uncaught TypeError: can't access property "setStyle", this.calmRoutesLayer is null
```

I don't think this is typically a problem on prod/staging because the calm routes layer is cached in those environments. Hence, this is a low priority bug, even if my root cause analysis is right.

Reproduce by hard refreshing and loading this route: http://localhost:8000/?sourceAddress=North+Branch+Trail%2C+Chicago%2C+IL+60630&targetAddress=Morton+Grove%2C+IL+60053&sourceCoordinates=41.9757289%2C-87.73300900000001&targetCoordinates=42.0405852%2C-87.78256209999999&showBbox=true

kalilsn:
Fixes #102

### DIFF
--- a/app/mbm/static/js/app.js
+++ b/app/mbm/static/js/app.js
@@ -518,7 +518,7 @@ export default class App {
     if (this.directionsRouteLayer) { this.map.removeLayer(this.directionsRouteLayer) }
     if (this.markers['source']) { this.map.removeLayer(this.markers['source']) }
     if (this.markers['target']) { this.map.removeLayer(this.markers['target']) }
-    this.calmRoutesLayer.setStyle({ opacity: 0.6 })
+    this.calmRoutesLayer?.setStyle({ opacity: 0.6 })
     this.hideRouteEstimate()
     this.sourceAddressString = ''
     this.targetAddressString = ''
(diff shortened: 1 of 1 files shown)


jeancochrane (approved):
Thanks for picking this up! If I'm understanding the code correctly, I think this is a reasonable 80/20 solution for the goal of preventing an error due to the race condition described in #102, but I don't think it resolves the underlying race condition. If the calm routes layer isn't initialized by the time the code tries to set its opacity to 30%, then the code will skip the opacity adjustment, and the calm routes layer will initialize with the default 60% opacity.

I'm fine merging this in to resolve the error, but I do think at some point we should fix the underlying race condition by waiting to set the 30% opacity style until the calm routes layer is loaded. Maybe we can just rescope #102 to reflect that final piece of work and bump it out of the current sprint, since I don't think the deeper fix is high priority. I'm curious what you think!

kalilsn:
Good point, I didn't quite realize what the purpose of that code was! I think my latest commit fully addresses the problem.

jeancochrane (approved):
Thanks for the deeper fix!

PR #108, opened 2026-05-30
Install js dependencies with npm and compile with esbuild

kalilsn:
This PR adds esbuild and moves away from including our dependencies in separate `<script>` tags and towards installing them with `npm`. There are still a few left in a vendor directory to avoid modifying the admin pages, but those should also be removed from the public facing app soon.

I made a variety of other changes too:

- Upgraded leaflet to 1.9.4, the latest 1.x.x release. v2 does not seem ready
- Replaced our leaflet google basemap plugin because it's deprecated. We're using https://gitlab.com/IvanSanchez/Leaflet.GridLayer.GoogleMutant/ which is what the author of the other plugin recommended and seems to improve the zooming experience on the map
- Replaced our leaflet-spin plugin with direct usage of spin.js
- Replaced the leaflet png marker with a fontawesome based one, since we only had a low res asset (now it's vector)
- Deleted some dead code and fixed a rare race condition on app startup

Most controversially, I also disabled our JS tests. Now that we're importing leaflet into the app, it's quite difficult to make them work because of all the mocks required. And since these tests don't actually interact with the html, they don't give me much confidence that the app works, since the JS is so tightly coupled to the html. Given that, I would like to replace these with some playwright based e2e tests that exercise the whole app from browser to database. As we refactor the frontend then I'd like to bring back in more js based unit tests.


app/Dockerfile
```diff
@@ -6,23 +6,24 @@ RUN curl -sL https://deb.nodesource.com/setup_22.x | bash -
...
 
 # Install Python requirements
 COPY ./requirements.txt /app/app/requirements.txt
 RUN pip install --no-cache-dir -r requirements.txt
 
 # Install Node requirements
-COPY ./package.json /app/app/package.json
+COPY ./package.json ./package-lock.json ./
+COPY ./patches ./patches
 RUN npm install
```
jeancochrane, line 18: **[Question, non-blocking]** When I was first testing this PR, I had to add the `--save-dev` flag here to install dev dependencies, or else `npm run build` would fail due to esbuild not being installed. However, the issue seems to have resolved itself on subsequent builds, so I'm not sure  if that's a necessary fix or if I was doing something wrong to start with. Do you have thoughts about whether we should need `--save-dev` in theory? For context, I was trying to keep all node dependencies isolated to the container, so I was following the frontend instructions that involve bringing up the `app` container and then running `docker compose exec app npm run dev` in a separate shell.

app/mbm/static/js/app.js
```diff
@@ -68,6 +71,26 @@ export default class App {
         autocomplete: null
       }
     }
+    this.spinner = new Spinner();
+
+    // Override the default marker icon with a fontawesome based icon
+    const iconHeight = 36
+    const iconWidth = 27
+    L.Marker.prototype.options.icon = L.divIcon({
+      html: '<i class="fas fa-map-marker-alt fa-3x" style="color: rgb(40, 132, 205);"></i>',
```
jeancochrane: **[Nitpick, non-blocking]** The only thing I don't love about this new fontawesome marker is that the hole in the marker is transparent, which means you can see the map underneath it:

Image

I don't think there's a super clean fix for this, since as far as I could tell from the [fontawesome docs](https://docs-v5.fontawesome.com/web/style), they don't provide non-transparent versions of the icons out of the box. The closest concepts seem to be "[masking](https://docs-v5.fontawesome.com/web/style/mask)" and "[duotone](https://docs-v5.fontawesome.com/web/style/duotone-icons)", neither of which seems exactly right; masking seems to be the inverse operation, and duotone might work but is exclusive to paid subscribers.

Two ideas I had for workarounds, neither of which strikes me as perfect:

1. Edit the SVG directly to fill in the hole, save the SVG as a file, and load it from a URL as a [custom icon](https://leafletjs.com/examples/custom-icons/)
    - Pro: Natively supported by Leaflet
    - Con: Loading an SVG icon from a URL feels like an unnecessary network request
 2. Use [fontawesome layers](https://docs.fontawesome.com/web/style/layer/) to layer a white circle underneath the map marker
     - Pro: Natively supported by fontawesome; implements solution directly in HTML/CSS, without requiring a network request to retrieve the icons
     - Con: Positioning the circle is a little tricky and I'm not confident how to do it in the most elegant way

Option 2 seems slightly preferable to me, so here's some code I tried and the resulting marker:

```suggestion
      html: `<span class="fa-layers fa-3x">
               <i class="fas fa-circle"
                  style="color: white; font-size: 12px; transform: translate(0, -4px)"></i>
               <i class="fas fa-map-marker-alt"
                  style="color: rgb(40, 132, 205);"></i>
             </span>`,
```

Image

app/mbm/static/js/app.js
```diff
@@ -68,6 +71,26 @@ export default class App {
...
+      popupAnchor: [0, -iconHeight],
+      className: 'defaultMarker'
+    });
+
+    if (document.readyState === "loading") {
+      // Start the app once the DOM is ready
+      document.addEventListener('DOMContentLoaded', this.start.bind(this))
+    } else {
+      this.start()
+    }
```
jeancochrane, line 97: **[Praise]** Nice catch here! To clarify my understanding: It seems like this is the block that fixes the startup race condition, in that it was previously possible for the script to execute _after_ the DOM had already loaded, in which case the script would never receive the expected `DOMContentLoaded` event. Let me know if that's incorrect!

app/mbm/static/js/app.js
```diff
@@ -172,14 +195,6 @@ export default class App {
       $(elem).data('state', state)
     }
 
-    const toggleControlElement = (elem, controlSelector) => {
-      if (isHidden(elem)) {
-        $(controlSelector).show()
-      } else {
-        $(controlSelector).hide()
-      }
-    }
```
jeancochrane, line 181: **[Praise]** Looks like this function was no longer used anywhere; if I'm right about that, thanks for deleting!

scripts/after_install.sh
```diff
@@ -34,6 +34,30 @@ sudo -H -u mbm $VENV_DIR/bin/pip install --upgrade "setuptools<50.0"
...
+    curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.40.4/install.sh | bash
+    # in lieu of restarting the shell
+    \. "$HOME/.nvm/nvm.sh"
+fi
+
+# Download and install Node.js:
+nvm install 24
+
+# Install and bundle js packages
+npm install
```
jeancochrane, line 56: **[Question, non-blocking]** I haven't tested the build, but I'm wondering if we need `--save-dev` here too, and if not, why not?

app/package-lock.json
```diff
@@ -1,4441 +1,1287 @@
...
-        "punycode": "1.3.2",
-        "querystring": "0.2.0"
+        "has-flag": "^4.0.0"
       },
+      "engines": {
+        "node": ">=8"
+      }
+    },
+    "node_modules/tmp": {
+      "version": "0.2.5",
```
jeancochrane: **[Nitpick, optional]** Quick flag that npm threw me security warnings for this version based on [a couple recent CVEs](https://github.com/raszi/node-tmp/security). Not sure if they affect us, so I leave it up to you to decide whether it's important to upgrade, particularly because this is an indirect dependency.

jeancochrane (approved):
Great work! The new frontend build pipeline is easy to use and I love the feel of the upgraded Google Maps integration on mobile.

The `package-lock.json` diff is huge and my comment is liable to get lost, so I want to make sure I add a quick pointer here to my comment on the `tmp` package security alert.

app/mbm/static/js/app.js
kalilsn: exactly right! i did manage to see this happen once locally and this code is pretty much exactly what mdn recommends for  `DOMContentLoaded` event listeners

app/mbm/static/js/app.js
kalilsn: yep, my editor indicated that to me

app/Dockerfile
kalilsn: AFAIK `--save-dev` is used when installing a single package to tell npm to put it in `devDependencies` (ie `npm install --save-dev esbuild`). just running `npm install` should install all dev dependencies as long as you don't have `NODE_ENV=production` or pass `npm install --production`.

I did test those instructions and they worked for me, but if you had issues we should amend them! I'll start by just adding `npm install` to the instructions for running entirely in docker.

app/package-lock.json
kalilsn: fixed!

scripts/after_install.sh
kalilsn: same as above! i can see how it's weird that we have packages like `esbuild` in `devDependencies` even though we're using them in prod. we could put everything in `dependencies` to reflect that. but i think we'll likely eventually move to bundling in ci (as is common) and then it will make more sense!

app/mbm/static/js/app.js
kalilsn: Applied this and made one small change to fix the positioning of the marker/tooltip since the shape changed. I think it looks good but you should probably test it out too! I'll deploy to staging

README.md
```diff
@@ -37,6 +37,21 @@ docker-compose up
...
+Then (in a separate terminal) make sure you're in the `app` directory, install the dependencies, and start the bundler in watch mode:
+```
+cd app
+npm install
+npm run dev
+```
+
+If you'd prefer to run the bundler in docker, you can do so inside the app container.
+```
+docker compose exec app npm install && npm run dev
```
jeancochrane: **[Nitpick, optional]** This command fails because bash splits the command on `&&` and interprets the second command as an isolated command to run once `docker compose exec app npm install` finishes executing:

```
$ docker compose exec app npm install && npm run dev

> mellow-bike-map@0.0.0 postinstall
> patch-package --patch-dir patches

patch-package 8.0.1
Applying patches...
leaflet.gridlayer.googlemutant@0.16.0 ✔

up to date, audited 65 packages in 1s

18 packages are looking for funding
  run `npm fund` for details

found 0 vulnerabilities
npm error code ENOENT
npm error syscall open
npm error path /Users/goobzie/code/mellow-bike-map/package.json
npm error errno -2
npm error enoent Could not read package.json: Error: ENOENT: no such file or directory, open '/Users/goobzie/code/mellow-bike-map/package.json'
npm error enoent This is related to npm not being able to find a file.
npm error enoent
```

One possible fix is to wrap the container's command in quotes and execute it with `bash -c`:

```suggestion
docker compose exec app bash -c "npm install && npm run dev"
```

jeancochrane (approved):
Awesome work! I'm excited for the mobile map to finally feel usable.

Tangentially related: This PR is reminding me that I've been meaning to switch us to hashed static files so that we can work around our users' browser caches when we deploy. I thought that hashing would need to be part of our frontend build pipeline, but it actually seems like Django's [`ManifestStaticFilesStorage` backend](https://docs.djangoproject.com/en/3.1/ref/contrib/staticfiles/#django.contrib.staticfiles.storage.ManifestStaticFilesStorage) will handle it for us with very few changes. I think I'll open a new issue for this, unless you feel strongly that we should handle it in the frontend build step.

README.md
kalilsn: oops, guess it just succeeded on my host machine when i tried it! thanks for checking

kalilsn:
> Awesome work! I'm excited for the mobile map to finally feel usable.

Thank you!! I'm still noticing some performance issues on mobile that I think we can alleviate by splitting the mellow route layer into tiles + defaulting to a bit more zoomed in view. Also interestingly, the specific zoom issue that was fixed by switching plugins is actually still present when I test it on macos firefox.

> Tangentially related: This PR is reminding me that I've been meaning to switch us to hashed static files so that we can work around our users' browser caches when we deploy. I thought that hashing would need to be part of our frontend build pipeline, but it actually seems like Django's [`ManifestStaticFilesStorage` backend](https://docs.djangoproject.com/en/3.1/ref/contrib/staticfiles/#django.contrib.staticfiles.storage.ManifestStaticFilesStorage) will handle it for us with very few changes. I think I'll open a new issue for this, unless you feel strongly that we should handle it in the frontend build step.

I'm fine letting django handle it!

PR #112, opened 2026-06-07
Update `mellowroute.json` fixture for Dec 2025 data, and move Halsted to Emerald/Union in Bridgeport

jeancochrane:
This PR makes two changes to the `mellowroute.json` fixture:

1. **Updates the fixture to include OSM IDs for ways as of December 2025** (the most recent data import). I was working on updating these tags in the [`drop-unimportant-sidewalks`](https://github.com/jeancochrane/mellow-bike-map/tree/jeancochrane/drop-unimportant-sidewalks) branch, and the new tags eventually got deployed to staging/prod, but I never got around to merging in the new fixture since #90 ended up superceding the `drop-unimportant-sidewalks` branch and I forgot to update the fixture in that PR. I committed this version of the fixture in bf6c17ad6c4c81d959c40453ba20abbc596dfc17 on this branch; the fixture in that commit should exactly match the current state of the prod and staging databases.

2. **Removes Halsted in Bridgeport from the `street` tag, and tags Emerald and Union instead**. I've been feeling wary of this stretch of Halsted for a while now, and [Riley O'Neil's tragic killing yesterday](https://www.chicagotribune.com/2026/06/06/bicyclist-killed-in-bridgeport-crash-was-cdot-employee-devoted-to-bike-infrastructure/) has left me convinced that Halsted is not safe for biking between Pershing and the river. I replaced Halsted with Emerald/Union in 8cef92de1a08018feec308a7ebca2907d2c64560.

Here is the current version of the map on prod:

Screenshot 2026-06-06 at 10 34 28 PM

Here is that same area, re-tagged with the changes from this branch (Emerald/Union for north/south travel through Bridgeport, rather than Halsted):

Screenshot 2026-06-06 at 10 36 13 PM

I double-checked Emerald and Union from 26th St to Root St on Street View to confirm that all intersections are either signalized, or include an all-way stop sign. I have been taking these two streets regularly from 35th to 26th for the past few months and I feel comfortable recommending them as chill routes. I haven't personally ridden them from 35th to Root yet, but Street View seems to show basically the same street quality as the northern section.

Note that while the South Halsted protected lane technically goes as far north as Pershing, I stopped tagging it and switched over to Emerald/Union at Exchange (which quickly turns into Root, visible on the map screenshots above). This is because I felt the rail bridge underpass at ~4000 South to be dicey when we biked under it last week. Root/Exchange also seems marginally better than Pershing for the one-or-two-block hop that is required to transfer from Halsted to Emerald/Union.

I did not tag Halsted as a `route` (orange on the map) even though it is technically a city-recommended route with painted lanes. I chose not to tag it because I dislike the `route` tag, find it to be counterproductive, and plan to remove it soon.


kalilsn (approved):
Agree with all of your thinking and IME these streets are chill.

One thought though: how about removing the mellow tag from 35th? I'm sure it's a commonly taken route because there aren't many other places to cross the river (which also means this shouldn't affect routing much) but I would really not describe any part of 35th as chill.

